PR #234, opened 2021-08-11
feat: scroll selected resource into view if not visible

catalinhoha:
introduced AppContext to avoid prop drilling of the windowSize

This PR...

## Changes

- scrolls selected resource into view if it's not visible, which allows us to click links in the navigator/editor and the newly selected resource will be scrolled into view

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
thanks @catalinhoha - does this also handle the case of expanding the containing section/subsection if needed to show the selected resource?

olensmar:
we can do it in a separate issue/PR if you prefer - i.e. "Auto-expand sections when scrolling into view"...

catalinhoha:
ok, that works, let's merge this in

PR #246, opened 2021-08-12
refactor: ResourceTypeHandler interface

olensmar:
For encapsulating logic specific to each k8s resource kind, this is a pre-cursor for external resourcetype plugins and internal refactoring for modularity. If approved the next step would be to:

- create a mechanism that adds available handlers to the global config
- use these handlers everywhere applicable
- refactor one resourcetype to use a handler instead of the existing "hardcoded" approach

once that works we can proceed and refactor for all resourcetypes.

(perhaps it should be called ResourceKindHandler?)


kambydyne (approved):

olensmar:
Looks great! Should we merge?

kambydyne (approved):

kambydyne:
Great work again, Catalin!

PR #250, opened 2021-08-15
feat: Add electronlogger to the project

erdkse:
This PR...

Will add `electron-logger` library to project. So in release mode, we will be able to see log messages of the application or We will be able to request log messages from end-users.

## Changes

- Added related library to the project
- Assigned log functions to console functions. So developers can continue to use log functions as it is.

## Fixes

-

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar (approved):
lgtm!

PR #261, opened 2021-08-17
Erdikose/feature/automated updates

erdkse:
This PR...

## Changes

the behaviour of update to the new version. Now the end-users will be able to see a button in the settings pane and they will be able to check and install the new versions of the application.

## Fixes

-

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/SettingsDrawer/SettingsDrawer.tsx
```diff
@@ -93,6 +99,33 @@ const SettingsDrawer = () => {
     }
   };
 
+  const checkUpdateAvailability = () => {
+    console.log('checkUpdateAvailability');
```
olensmar, line 103: do we want to keep all these log statements?

electron/main.ts
```diff
@@ -58,6 +61,18 @@ ipcMain.on('run-helm', (event, args: any) => {
   }
 });
 
+ipcMain.on('app-version', event => {
+  event.sender.send('app-version', {version: app.getVersion()});
+});
+
+ipcMain.on('check-update-available', () => {
+  autoUpdater.checkForUpdatesAndNotify();
```
olensmar, line 69: how/where does the autoUpdater actually check for a new version?

src/components/organisms/SettingsDrawer/SettingsDrawer.tsx
erdkse: No, before I change draft PR to a real one I will delete all logs.

electron/main.ts
erdkse: It checks our Github releases. If there is a new release It notifies the application and downloads it.

olensmar: ok thanks - is this configurable? i.e. can we tell it check a specific URL? If so - what do we have to host at that URL?

erdkse: Yeah, it is configurable. All we need to do is to place our files on the server. Supported providers are `GitHub Releases`, `Amazon S3`, `DigitalOcean Spaces`, `Bintray`. We can also use our custom server. But I think after Github Releases the best option is AWS S3 with some authentication gate. It works well with private Github repos as well. We just need to set Github Access Token to our environments.

olensmar: ok thanks - we can start with GitHub for now - what I'm after is getting analytics on how often this is checked - as an initial indicator of monokle usage.. i'll create a separate ticket for moving this somewhere else, there might be other aspects we need to be mindful of...

erdkse: We can track this event on the client side. No server-side implementation is needed for this purpose I think. Even google analytics will do what we want. We can get information per user.

src/components/organisms/PageFooter/PageFooter.tsx
```diff
@@ -18,15 +18,26 @@ const StyledFooter = styled(Footer)`
...
-  }`;
+  ipcRenderer.send('app-version');
+  ipcRenderer.once('app-version', (_, {version}) => {
+    setAppVersion(version);
+  });
+
+  useEffect(() => {
+    setFooterText(
+      `Monokle ${appVersion} - kubeshop.io 2021${
+        rootEntry && rootEntry.children ? ` - ${rootEntry.filePath} - ${nrOfFiles} files` : ''
```
olensmar, line 37: could we show in the footer if Monokle is downloading an update - and/or when a download has finished so the user knows they can install it?

olensmar (approved):
LGTM - updating the page footer is a nice have...

src/components/organisms/PageFooter/PageFooter.tsx
erdkse: When there is an available update I show a notification.
Screen Shot 2021-08-24 at 13 54 48

And if the update is installing I show it inside the settings drawer.
Screen Shot 2021-08-24 at 13 54 57

And When the update is ready for install. The user will see this button.
Screen Shot 2021-08-24 at 13 55 07


Do you still think we need to show in the footer. I don't want to show a lot of notifications to users.

kambydyne (approved):
lgtm, let's see how it plays out with the new signed installers.

PR #272, opened 2021-08-18
feat: Add command installer & manage to run application via CLI

erdkse:
This PR...

## Changes

Will allow users to run application via terminal with specified resources.

## Fixes

-

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test


cli/terminal.ts
```diff
@@ -0,0 +1,51 @@
...
+const path = require('path');
+const fs = require('fs');
+
+const installCommand = (commandPath: string, commandName: string): Promise<void> => {
+  return new Promise((resolve, reject) => {
+    if (typeof commandPath !== 'string' || typeof commandName !== 'string') {
+      reject(new TypeError('Expected a string'));
+    }
+
+    if (process.platform !== 'darwin') {
```
olensmar, line 11: can we make this work on other platforms (i.e. windows)? or would that require a different approach?

electron/main.ts
```diff
@@ -141,3 +143,25 @@ app.whenReady().then(() => {
     }
   });
 });
+
+const args = parseArgs(process.argv.slice(1));
```
olensmar, line 147: we should support the passing of a folder when launching, which would automatically  open that folder (i.e. setRootFolder..) in the UI.. i.e. as a user I should be able to type

`monokle . `

or 

`monokle manifests`

cli/terminal.ts
erdkse: The current approach will work only on macOS. It will require a different approach to support Windows. First, I am planning to support macOS and make everything working well. I am actively working on this feature right now. I may end today by supporting Windows too.

electron/main.ts
erdkse: yeah, I can get it right now but I didn't send it yet. I am reading the code first. I need to figure out how we load a directory. As you see I get the parameter of `executed-from`. I am planning to send it inside react and handle it there.

Screen Shot 2021-08-19 at 10 54 51

olensmar: ok great !

PR #286, opened 2021-08-24
feat: added functionaly to apply files when they are selected

olensmar:
## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/ActionsPane/ActionsPane.tsx
+++ b/src/components/organisms/ActionsPane/ActionsPane.tsx
@@ -19,10 +19,11 @@ import {K8sResource} from '@models/k8sresource';
 import {isKustomizationResource} from '@redux/services/kustomize';
 import {FileMapType, ResourceMapType} from '@models/appstate';
 import {ThunkDispatch} from 'redux-thunk';
-import {ApplyTooltip, DiffTooltip} from '@constants/tooltips';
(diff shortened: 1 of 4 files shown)


src/components/organisms/ActionsPane/ActionsPane.tsx
```diff
@@ -131,6 +154,8 @@ const ActionsPane = (props: {contentHeight: string}) => {
   const applySelectedResource = useCallback(() => {
```
catalinhoha, line 154: Might need to find a better name for this function and add `selectedPath` to the useCallback deps

src/redux/thunks/applyFile.ts
```diff
@@ -0,0 +1,78 @@
...
+import {getShellPath} from '@utils/shell';
+import {setApplyingResource} from '@redux/reducers/main';
+import fs from 'fs';
+import {getAbsoluteFileEntryPath} from '@redux/services/fileEntry';
+
+/**
+ * Invokes kubectl for the content of the specified resource
+ */
+
+function applyFileToCluster(filePath: string, kubeconfig: string) {
```
catalinhoha, line 16: I think a better name for this would be `spawnProcessToApplyFile` or `spawnApplyFileProcess`

PR #292, opened 2021-08-25
Erdikose/feature/refresh cluster resources

erdkse:
This PR...

## Changes

Reload cluster objects options is added. After the first load, the button turns `Reload Cluster objects` and reads the same config file again.

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


erdkse:
Ready to re-review @olensmar @catalinhoha

src/components/organisms/ClustersPane/ClustersPane.tsx
```diff
@@ -97,10 +114,9 @@ const ClustersPane = () => {
               type="primary"
               ghost
               loading={previewType === 'cluster' && previewLoader.isLoading}
```
olensmar, line 116: for some reason the loading animation is rendered over the label - not next to it!?

erdkse: Hmm, I saw as next to it. Let me check!

erdkse: I always see it next to the label. Can you please send me a ss?

olensmar: image
you can see the spinning circle over the initial 'L' - this is only temporary - after a second or two the label moves to the right and the button looks ok...

olensmar (approved):

src/components/organisms/ClustersPane/ClustersPane.tsx
catalinhoha: I can confirm this, it happens because the UI freezes while the preview is loading. Probably that's why the animation on Skeletons doesn't work either..

olensmar: ok thanks @catalinhoha - then let's live with it for now and consider moving the loading of cluster resources to the main thread in a separate issue!

PR #305, opened 2021-08-29
fix: Fix resource count update on file change

erdkse:
This PR...

## Changes

*

## Fixes

* Updating the resource count of a file when changing it.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/FileTreePane/FileTreePane.tsx
+++ b/src/components/organisms/FileTreePane/FileTreePane.tsx
@@ -253,7 +253,7 @@ const FileTreePane = () => {
       setExpandedKeys(Object.keys(fileMap).filter(key => fileMap[key]?.children?.length));
       shouldExpandAllNodes.current = false;
     }
-  }, [fileMap]);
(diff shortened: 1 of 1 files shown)


src/components/organisms/FileTreePane/FileTreePane.tsx
```diff
@@ -253,7 +253,7 @@ const FileTreePane = () => {
       setExpandedKeys(Object.keys(fileMap).filter(key => fileMap[key]?.children?.length));
       shouldExpandAllNodes.current = false;
     }
-  }, [fileMap]);
+  }, [resourceMap]);
```
olensmar, line 256: not both fileMap and resourceMap?

erdkse: I debugged as when we change the `fileMap` we also update the `resourceMap` so I thought there is no need to add fileMap into dependecies.

PR #314, opened 2021-08-31
fix: allowed editing/reload of previews

olensmar:
This PR...

## Changes

- makes it possible to edit kustomization/values files when they are being previewed, and adds a "reload" action to reload the preview without having to exit it first. This makes the kustomization/helm debugging workflow smoother (less clicks..)

## Fixes

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/molecules/Monaco/Monaco.tsx
```diff
@@ -250,9 +253,18 @@ const Monaco = (props: {editorHeight: string}) => {
     setDirty(false);
   }, [fileMap, selectedPath, selectedResourceId, resourceMap]);
 
+  // read-only if we're in preview mode and another resource is selected - or if nothing is selected at all
+  function isReadOnlyMode() {
```
catalinhoha, line 257: this could be wrapped in a `useCallback`

catalinhoha (approved):

kambydyne (approved):

PR #316, opened 2021-08-31
Erdikose/fix/source code selection

erdkse:
This PR...

## Changes

-

## Fixes

- prevents complete source code selection on first resource selection

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/Monaco/Monaco.tsx
+++ b/src/components/molecules/Monaco/Monaco.tsx
@@ -79,6 +79,7 @@ const Monaco = (props: {editorHeight: string}) => {
   const [isDirty, setDirty] = useState(false);
   const [hasWarnings, setWarnings] = useState(false);
   const [isValid, setValid] = useState(true);
+  const [firstCodeLoadedOnEditor, setFirstCodeLoadedOnEditor] = useState(false);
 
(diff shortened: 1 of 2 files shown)


src/components/molecules/Monaco/Monaco.tsx
```diff
@@ -329,11 +329,11 @@ const Monaco = (props: {editorHeight: string}) => {
   }, [selectedResourceId, resourceMap]);
 
   useEffect(() => {
-    if (editor) {
+    if (editor && code) {
```
olensmar, line 332: hm... isn't this going to set the selection every time the code changes (which is after each debounce) ? What happens if you are editing in a scrolled-down portion of the document?

erdkse: hmm, yeah! You are right I didn't think and test with the case you say.

catalinhoha: I think we need logic to detect when code changes from empty string for the first time and set the selection at the top, but not on every code change

erdkse: I did the similar approach

olensmar (approved):

PR #320, opened 2021-08-31
Steen/feat/sign binaries

kambydyne:
This PR...

## Changes

- Adds signing and notarization for macos builds

## Fixes

-

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test


package.json
```diff
@@ -139,7 +139,8 @@
     },
     "afterSign": "./notarization/afterSignHook.js",
     "dmg": {
-      "icon": "build/icon.png"
+      "icon": "build/icon.png",
+      "writeUpdateInfo": false
```
erdkse, line 143: Do we need to add this to disable blockmaps? This may harm automatic updates if that doesn't produce update.yml file.

package.json
```diff
@@ -149,7 +150,8 @@
       "oneClick": true,
       "installerIcon": "build/icon.ico",
       "uninstallerIcon": "build/icon.ico",
-      "installerHeaderIcon": "build/icon.ico"
+      "installerHeaderIcon": "build/icon.ico",
+      "differentialPackage": false
```
erdkse, line 154: same..

package.json
kambydyne: The latest.yml still exists. Let's see if it works, otherwise we can remove those entries again.

PR #353, opened 2021-09-09
Delete resource from file/cluster

catalinhoha:
This PR...

## Changes

- add logic to delete a resource from a file or from the cluster
- show ActionsMenu on hover over resources in navigator pane
- show delete resource action with confirm modal

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/NavigatorRowLabel/NavigatorRowLabel.tsx
+++ b/src/components/molecules/NavigatorRowLabel/NavigatorRowLabel.tsx
@@ -1,8 +1,10 @@
 import React, {useState, useEffect, useContext, useCallback} from 'react';
-import {Popover, Typography, Divider} from 'antd';
+import {Popover, Typography, Divider, Dropdown} from 'antd';
 import styled from 'styled-components';
+import {useSelector} from 'react-redux';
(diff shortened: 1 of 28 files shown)


kambydyne (approved):

src/components/molecules/NavigatorRowLabel/NavigatorRowLabel.tsx
```diff
@@ -350,6 +361,13 @@ const NavigatorRowLabel = (props: NavigatorRowLabelProps) => {
           </StyledIconsContainer>
         </Popover>
       )}
+      {isHovered && (!isInPreviewMode || (isInPreviewMode && previewType === 'cluster')) && (
```
olensmar, line 364: perhaps we should always show the menu and enable/disable menu actions depending on which state we're in? (since different actions might apply to different types of resources and/or previewModes)

catalinhoha: sure, I will do that in a following PR with either clone or rename resource

PR #357, opened 2021-09-10
refactor: disable selecting non-yaml files

catalinhoha:
This PR...

## Changes

- disables non-yaml file selections

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/FileTreePane/FileTreePane.tsx
```diff
@@ -309,6 +309,9 @@ const FileTreePane = () => {
   }, [selectedPath]);
 
   const onSelect = (selectedKeysValue: React.Key[], info: any) => {
+    if (!['.yaml', '.yml'].includes(path.extname(info.node.key))) {
```
olensmar, line 312: you could check against the appConfig.fileIncludes array?

catalinhoha: but then the user could include other file types that can be selected and shown in the editor, do we want that for now?

olensmar: those files would be parsed for manifests - so I think it's ok to make them editable also

catalinhoha: I see, okay, will modify it

olensmar (approved):

PR #390, opened 2021-09-22
fix: Add env.ts to handle environment variables

erdkse:
This PR...

## Changes

- The way of handling environment variables

## Fixes

- NODE_ENV undefined issue

## How to test it

- Copy env.example.ts file as env.ts and set env as production and get the build.

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1,6 +1,9 @@
+/* eslint-disable import/order */
 /* eslint-disable import/first */
 import moduleAlias from 'module-alias';
+import * as ElectronLog from 'electron-log';
 
+Object.assign(console, ElectronLog.functions);
 moduleAlias.addAliases({
   '@constants': `${__dirname}/../src/constants`,
   '@models': `${__dirname}/../src/models`,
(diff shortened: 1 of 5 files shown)


olensmar:
@erdkse does this require any changes to the nightly/release build workflows we have on GitHub?

erdkse:
> @erdkse does this require any changes to the nightly/release build workflows we have on GitHub?

Yes, we need to set env variable in env.ts file. Otherwise we can't have build.

olensmar:
> > @erdkse does this require any changes to the nightly/release build workflows we have on GitHub?
> 
> Yes, we need to set env variable in env.ts file. Otherwise we can't have build.

ok thanks - can you make sure this is done for our nightly builds?

olensmar:
sorry for punting on this @erdkse  - i would just like to understand why this is even necessary - since the electron-is-dev solution worked fine in the 1.1.0 release - something must have broken on the way...

erdkse:
Yeah, you are right. `electron-is-dev` library reads ELECTRON_IS_DEV property in `process.env`. But currently, we can't set any value to process.env now. So I made this PR to allow us to continue. If we have time I can investigate what is going on.

olensmar (approved):

electron/main.ts
```diff
@@ -10,12 +13,18 @@ moduleAlias.addAliases({
   '@root': `${__dirname}/../`,
 });
 
+import ENV from '../env';
+
+if (!ENV.env || ENV.env === 'development') {
```
olensmar, line 18: should the logic perhaps be the other way around - i.e. production is default if no file exists? in which case our builds wouldn't have to be updated?

erdkse: Yes, makes sense. I will change in that way.

kambydyne:
I'm merging this in, so we can test it for the release

PR #583, opened 2021-11-03
feat: added a possibility to copy any text in the application

Toxer10:
## Changes

- Added a possibility to copy any text in the application.
- Added copy-to-clipboard functionality in notification.

## Fixes

- Changed the behavior when the user could not copy any text in the application because this feature was working only under specific conditions.

### DIFF
--- a/electron/menu.ts
+++ b/electron/menu.ts
@@ -141,17 +141,22 @@ const fileMenu = (store: any): MenuItemConstructorOptions => {
 const editMenu = (store: any): MenuItemConstructorOptions => {
   const mainState: AppState = store.getState().main;
   const uiState: UiState = store.getState().ui;
+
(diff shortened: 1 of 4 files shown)


erdkse:
Just small additions;

Maybe instead of copying just the message property. We can generate a format for the title and the message together and copy it instead.

Shall we add the copy icon to the toast message template too? Wdyt?

Other than that looks great.

Toxer10:
> Just small additions;
> 
> * Maybe instead of copying just the message property. We can generate a format for the title and the message together and copy it instead.
> * Shall we add the copy functionality to the toast message template too? Wdyt?
> 
> Other than that looks great.

- Can create something like this - Title: Folder Import. Description: 0 resources found in 1 files. 
- Let's hear the opinions of others regarding the second thing.

erdkse (approved):

PR #594, opened 2021-11-04
feat: displayed the number of filters applied

Toxer10:
## Changes

- Displayed the number of filters applied.
- Refactored code a bit: type and interface declarations.
- Added new reusable component IconWithPopover.

## Fixes

- Fixed behavior when user could not reset all the filters.

### DIFF
--- a/src/components/molecules/ResourceFilter/ResourceFilter.tsx
+++ b/src/components/molecules/ResourceFilter/ResourceFilter.tsx
@@ -63,11 +63,9 @@ const makeKeyValuesFromObjectList = (objectList: any[], getNestedObject: (curren
 
 const ResourceFilter = () => {
   const dispatch = useAppDispatch();
-
-  // TODO: Should we use here Antd Form instead of storing all the field values in useStates?
-  const [name, setName] = useState('');
-  const [kind, setKind] = useState(ALL_OPTIONS);
-  const [namespace, setNamespace] = useState(ALL_OPTIONS);
+  const [name, setName] = useState<string>();
+  const [kind, setKind] = useState<string>();
+  const [namespace, setNamespace] = useState<string>();
   const [labels, setLabels] = useState<Record<string, string | null>>({});
   const [annotations, setAnnotations] = useState<Record<string, string | null>>({});
   const allNamespaces = useNamespaces({extra: ['all', 'default']});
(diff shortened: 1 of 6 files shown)


src/components/molecules/ResourceFilter/ResourceFilter.tsx
```diff
@@ -88,27 +90,27 @@ const ResourceFilter = () => {
...
     setAnnotations({});
   };
 
   const updateKind = (selectedKind: string) => {
-    if (selectedKind === ALL_OPTIONS) {
-      setKind(undefined);
-    } else {
-      setKind(selectedKind);
-    }
+    // if (selectedKind === ALL_OPTIONS) {
```
Toxer10, line 101: Wanted to ask why is this implemented in this way? What is the reason?

catalinhoha: It was needed to show the string `<all>` in the `Select` when the value is `undefined`

src/components/organisms/NavigatorPane/NavigatorPane.tsx
```diff
@@ -18,16 +19,27 @@ import SectionRenderer from './SectionRenderer';
...
   const fileMap = useAppSelector(state => state.main.fileMap);
+  // TODO: remove this any
+  const resourceFilters: any = useAppSelector(state => state.main.resourceFilter);
+
   const isInClusterMode = useSelector(isInClusterModeSelector);
   const isInPreviewMode = useSelector(isInPreviewModeSelector);
 
+  const appliedFilters = useMemo(() => {
+    return Object.keys(resourceFilters)
+      .map(filterName => {
```
catalinhoha, line 37: You could use `Object.entries()` instead of `.keys` and then `.map`

src/components/organisms/NavigatorPane/NavigatorPane.tsx
```diff
@@ -18,16 +19,27 @@ import SectionRenderer from './SectionRenderer';
...
 
-const NavPane = () => {
+const NavPane: React.FC = () => {
   const dispatch = useAppDispatch();
   const {windowSize} = useContext(AppContext);
   const windowHeight = windowSize.height;
   const navigatorHeight = windowHeight - NAVIGATOR_HEIGHT_OFFSET;
 
   const fileMap = useAppSelector(state => state.main.fileMap);
+  // TODO: remove this any
```
catalinhoha, line 29: Agreed, `any` can be replaced with `ResourceFilterType` from `models/appstate`

src/components/organisms/NavigatorPane/NavigatorPane.tsx
Toxer10: Thanks for the advice!

src/components/organisms/NavigatorPane/NavigatorPane.tsx
Toxer10: Thanks for the advice!

catalinhoha (approved):

PR #603, opened 2021-11-08
feat: added expand/collapse buttons in file explorer tree

Toxer10:
## Changes

- Added expand/collapse buttons


olensmar:
looks like the icon is getting on the wrong side of the divider when cramped - can we fix this?

Toxer10:
image
It happens in the middle too. We should change the layout to prevent this behavior on all sides: left, middle, and right.

olensmar:
Yes - please look into this!

olensmar (approved):
lgtm!

PR #604, opened 2021-11-08
feat: auto import sorting

Toxer10:
This PR...

## Changes

- Added auto import sorting. To discuss with the team


erdkse:
Do we have to install an extension for this?

Toxer10:
Yes. I found one for VS Code called TypeScript Import Sorter https://marketplace.visualstudio.com/items?itemName=mike-co.import-sorter. Do we need the one which works not only in VS Code?

olensmar:
@Toxer10 we need one that works with es-lint or something similar - it should be automated as a pre-commit hook and  shouldn't require any specific editor

olensmar (approved):
LGTM

PR #673, opened 2021-11-16
fix: intermediate fix that ignores lastProbeTime/creationTimestamp validation errors

### LINKED ISSUE #668: Option to ignore certain fields when validating resources
Certain resource properties often contain an invalid value when retrieved from a cluster - for example creationTimestamp in the screenshot below:

image

To avoid this error cluttering up the interface and "hiding" any real errors it would be helpful if Monokle could maintain a list of paths to ignore during validation

olensmar:
I suggest we get this in until we implement fix #668


devcatalin (approved):
lgtm!

PR #675, opened 2021-11-16
fix: reprocess all resources for file updates

olensmar:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/redux/services/resourceRefs.ts
+++ b/src/redux/services/resourceRefs.ts
@@ -1,5 +1,4 @@
 import {isKustomizationPatch, isKustomizationResource} from '@redux/services/kustomize';
-import {Draft} from '@reduxjs/toolkit';
 
 import {ResourceMapType, ResourceRefsProcessingOptions} from '@models/appstate';
 import {K8sResource, RefNode, RefPosition, ResourceRef, ResourceRefType} from '@models/k8sresource';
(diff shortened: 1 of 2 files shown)


src/redux/services/resourceRefs.ts
```diff
@@ -510,3 +510,30 @@ export function processRefs(
...
+  // all existing resources with incoming refs to the resource (since they might have broken)
+  resource.refs
+    ?.filter(ref => isIncomingRef(ref.type))
+    .forEach(ref => {
+      if (ref.target?.type === 'resource' && ref.target.resourceId) {
+        resources.push(ref.target.resourceId);
+      }
+    });
+
+  // all existing resources with broken refs (since they might now work) -> this should be improved to
```
devcatalin, line 531: good point, should we refactor `reprocessResources` from `services/resource` to do a filtering like this instead of checking of for all dependent resource kinds and then reprocessing all resources of that kind?

olensmar: i was thinking we would add an optional list of targetResourceKinds to the unsatisfied ref (since we know them when we try to discover the ref) - which we could use here for further filtering!?

olensmar: actually - looking at the code it seems we already add this to the ref - will see how that works!

devcatalin (approved):

src/redux/services/resourceRefs.ts
olensmar: no - this requires more work... will merge this and tackle that separately...

Issue #641, opened 2021-11-11
Retrieve unhandled resource kinds from cluster


olensmar:
Monokle currently only retrieves resource kinds it knows if from the selected context/cluster - we should retrieve/show all resources as we do for files - especially in regard to the Cluster Comparison which currently won't support any unkown resource kinds.

PR #681, opened 2021-11-16
fix: added retrieval of custom resource objects from cluster

olensmar:
This PR implements #641 

## How to test it

- retrieve resources from a cluster containing custom resources - they should show up at the bottom of the navigator under an "Unknown" section

## Screenshots

image

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/redux/thunks/previewCluster.ts
+++ b/src/redux/thunks/previewCluster.ts
@@ -1,15 +1,19 @@
 import {SetPreviewDataPayload} from '@redux/reducers/main';
+import {extractK8sResources, processParsedResources} from '@redux/services/resource';
 import {AppDispatch, RootState} from '@redux/store';
 import {createPreviewResult, createRejectionWithAlert, getK8sObjectsAsYaml} from '@redux/thunks/utils';
 import {createAsyncThunk} from '@reduxjs/toolkit';
 
 import {AlertEnum} from '@models/alert';
+import {ResourceRefsProcessingOptions} from '@models/appstate';
+import {K8sResource} from '@models/k8sresource';
 
-import {YAML_DOCUMENT_DELIMITER_NEW_LINE} from '@constants/constants';
+import {PREVIEW_PREFIX, YAML_DOCUMENT_DELIMITER_NEW_LINE} from '@constants/constants';
 
 import {ResourceKindHandlers} from '@src/kindhandlers';
 
 import * as k8s from '@kubernetes/client-node';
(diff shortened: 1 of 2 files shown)


src/redux/thunks/previewCluster.ts
```diff
@@ -25,7 +29,7 @@ const previewClusterHandler = async (configPath: string, thunkAPI: any) => {
           .then(items => getK8sObjectsAsYaml(items, resourceKindHandler.kind, resourceKindHandler.clusterApiVersion))
       )
     ).then(
-      results => {
+      async results => {
```
erdkse, line 32: Why not 

```
const results = await Promise.allSettled...
(inside of then)
```

src/redux/thunks/previewCluster.ts
```diff
@@ -95,3 +113,108 @@ export const repreviewCluster = createAsyncThunk<
...
+          // @ts-ignore
+          getK8sObjectsAsYaml(response.body.items)
+        )
+    );
+
+    if (!previewResult.previewResources) {
+      previewResult.previewResources = {};
+    }
+
+    await Promise.allSettled(customObjects).then(
```
erdkse, line 195: No need to use `then` if it is async/await

```
const customResults = await Promise.allSettled(customObjects);
(inside of then)
```

erdkse (approved):
lgtm

src/redux/thunks/previewCluster.ts
olensmar: thanks @erdkse  - not sure what you mean - allSettled returns a promise  - it would still have to be 

```
const customResultsPromise = Promise.allSettled(customObjects);
customResultsPromise.then(
  customResults => { .. },
  reason => { ... }
);
```
!?

olensmar: I'll try that - thanks!

devcatalin: Yes, the alternative would be:
```try {
  const customResults = await Promise.allSettled(customObjects);
  // do something with customResults...
} catch(reason: any) {
}```

olensmar:
@erdkse i've made changes in line with your suggestions - please have a look and let me know if this is what you had in mind..

erdkse:
Thank you Ole, looks perfect!

erdkse (approved):

PR #745, opened 2021-11-22
fix: improved action propagation across threads and fileMonitor/times…

olensmar:
…tamp handling

This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/FileTreePane/FileTreePane.tsx
```diff
@@ -668,6 +668,15 @@ const FileTreePane = () => {
     [loadLastFolderOnStartup, setFolder, recentFolders]
   );
 
+  // called from main thread to avoid running setRootFolder in main thread
+  useEffect(() => {
+    ipcRenderer.on('set-root-folder', (_, data) => {
```
devcatalin, line 673: please add logic to remove the event listener using `return` like we did on the `useEffect` below this one to avoid memory leaks

src/redux/services/fileMonitor.ts
```diff
@@ -15,55 +15,57 @@ let watcher: FSWatcher;
...
-  }
-
-  watcher = watch(folder, {
-    ignored: appConfig.scanExcludes,
-    ignoreInitial: true,
-    persistent: true,
-    usePolling: true,
-    interval: 1000,
-  });
+    const watched = Object.keys(watcher.getWatched());
```
devcatalin, line 18: here if the watcher exists we should update the `ignored` property to the most up to date `appConfig.scanExcludes`

olensmar: good catch - unfortunately there doesn't seem to be a way to change the ignored property after creation - i'll go back to recreating the watcher for now

devcatalin (approved):
lgtm!

PR #755, opened 2021-11-23
fix: hide dependency warning if the user can run kubectl kustomize

Toxer10:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -27,11 +27,6 @@ export const checkMissingDependencies = (dependencies: Array<string>): Array<str
       });
       return false;
     } catch (e: any) {
-      // If kubectl kustomize was ran and threw an exception that it could not find any kustomization file - should return false
-      if (e.message.includes("unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization'")) {
-        return false;
-      }
-
       return true;
     }
   });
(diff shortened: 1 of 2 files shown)


src/utils/index.ts
```diff
@@ -26,7 +26,12 @@ export const checkMissingDependencies = (dependencies: Array<string>): Array<str
         },
       });
       return false;
-    } catch (e) {
+    } catch (e: any) {
+      // If kubectl kustomize was ran and threw an exception that it could not find any kustomization file - should return false
+      if (e.message.includes("unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization'")) {
```
devcatalin, line 31: I think you can remove this and run `kubectl kustomize --help` instead so you get an error only if it's not recognized

devcatalin (changes requested):

src/utils/index.ts
Toxer10: Thanks for the advice!

Toxer10: Done

devcatalin (approved):
lgtm!

PR #779, opened 2021-11-25
feat: added an ability to save the file right after creating

Toxer10:
## Changes

- added an ability to save the file right after creating

### DIFF
--- a/src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
+++ b/src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
@@ -170,7 +170,7 @@ const NewResourceWizard = () => {
   const foldersList = useMemo(
     () =>
       Object.entries(fileMap)
-        .map(([key, value]) => ({folderName: key, isFolder: Boolean(value.children)}))
+        .map(([key, value]) => ({folderName: key.replace(path.sep, ''), isFolder: Boolean(value.children)}))
         .filter(file => file.isFolder),
     [fileMap]
   );
--- a/src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
+++ b/src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
@@ -1,6 +1,6 @@
 import {useEffect, useRef, useState} from 'react';
 
-import {Form, Input, Modal, Select} from 'antd';
+import {Checkbox, Form, Input, Modal, Select} from 'antd';
 
 import {InfoCircleOutlined} from '@ant-design/icons';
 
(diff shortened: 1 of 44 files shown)


olensmar:
thanks @EduardIvanov22 - i suggest the following changes:
- change the label to the checkbox to  "Save to folder {currentFolder}" - and automatically save to that folder (no prompting)
- currentFolder defaults to the current root folder
- add a "Change folder" button (after the label?) that the user can select to change the folder (which would open the dialog you already have) - the label would change accordingly once selected
- remember the last selected folder (if the user changes it from the default root folder)

what do you think? @devcatalin ? @erdkse ?

devcatalin:
I agree with all the points @olensmar mentioned.

For implementation I would just use the `createUnsavedResource` and then `saveUnsavedResource` giving it the path of the selected folder

olensmar:
thanks @devcatalin - @EduardIvanov22 you'll need to merge main into your branch - i recently made some changes to the dialogs and saveUnsavedResource action

Toxer10:
image
What do you think of this approach? @olensmar @devcatalin

olensmar:
looks good - the drop-down would contain all folders under the current root folder - correct? (with their relative path to the root folder)

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -103,26 +115,74 @@ const NewResourceWizard = () => {
...
+        apiVersion: formValues.apiVersion,
       },
       dispatch,
       jsonTemplate
     );
 
+    const fullFileName = getFullFileName(formValues.name);
+
+    if (shouldSaveToFolder) {
+      setTimeout(() => {
```
olensmar, line 154: could you simply add an optional savePath property to the createUnsavedResource method above - which would save the resource after creating it? (if the path was provided)

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -103,26 +115,74 @@ const NewResourceWizard = () => {
...
+      Object.entries(fileMap)
+        .map(([key, value]) => ({folderName: key, isFolder: Boolean(value.children)}))
+        .filter(file => file.isFolder),
+    [fileMap]
+  );
+
+  const renderSelectOptions = () => {
+    return foldersList.map(folder => (
+      <Option key={folder.folderName} value={folder.folderName}>
+        {folder.folderName}
```
olensmar, line 181: suggest to remove the leading '/'

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
Toxer10: Sure

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
Toxer10: Sure

Toxer10: Done

olensmar (approved):

PR #788, opened 2021-11-25
feat: add file option to context menu

erdkse:
This PR...

## Changes

- Add create file option to context menu
- Add `create resource` option to context menu 
- Save resource instantly

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
thanks @erdkse - i'm not sure why we would have this since Monokle only handles manifest files - wouldn't it be better to have a "New Resource" menu item which starts the new-resource wizard with the selected folder preselected? (after we merge Eduards ongoing PR)

PR #791, opened 2021-11-26
fix: fixed wrong styling of context menu in k8s resources tab

Toxer10:
## Fixes

- fixed wrong styling of context menu in k8s resources tab

If we provide the overlay menu as another component e.g. ResourceActionsMenu, Ant Design does not recognize this component as a Menu and does not provide any Menu styles to it.


Toxer10:
image

olensmar (approved):

Issue #781, opened 2021-11-25
remove resource from editor when it is deleted via external tool


olensmar:
- open a resource  in the source editor
- delete the file containing the resource using an external tool
-> the file and resource are removed from the navigators but stays in the editor - should be removed there also...

PR #793, opened 2021-11-26
fix: improved processing of removed files and removed unused reducers

olensmar:
This PR...

## Changes

- #781 

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/redux/services/fileEntry.ts
```diff
@@ -503,7 +503,13 @@ export function removePath(absolutePath: string, state: AppState, fileEntry: Fil
     }
   }
 
-  reprocessResources([], state.resourceMap, state.fileMap, state.resourceRefsProcessingOptions, {
-    resourceKinds: removalSideEffect.removedResources.map(r => r.kind),
-  });
+  reprocessResources(
+    Object.values(state.resourceMap).map(r => r.id),
```
devcatalin, line 507: I'm wondering if we should filter out the preview resources from here since this gets called when a path is removed and that shouldn't affect any of the preview modes.

olensmar: makes sense - i'll improve accordingly

olensmar: fixed..

devcatalin (approved):

Issue #444, opened 2021-10-01
Filter from editor


olensmar:
When viewing a resource with some annotations I would like to immediately create a filter for a specific annotation value to see all other resources with the same value. Now I have to add this filter "manually" - I would like a popup-action allowing me to click on the value and select either "Create new filter" or "Add to filter..." which would automatically create and/or add the value to the current filter. This should apply to the following properties/values that we can currently filter on:
- namespaces
- annotations
- labels

For example when viewing the below resource I want to be able to click on one of the label values and add that to my current filter.

![image](https://user-images.githubusercontent.com/1917063/135591520-8af8133e-ee4d-4f18-afbf-fdebc41be62b.png)

PR #801, opened 2021-11-29
fix: added hover links to add filters in source editor

olensmar:
This PR adds hover actions to editor for extending the current resource filter - see #444 

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/Monaco/useCodeIntel.ts
+++ b/src/components/molecules/Monaco/useCodeIntel.ts
@@ -57,7 +57,7 @@ function useCodeIntel(
   const debouncedUpdate = debounce(() => {
     clearCodeIntel();
     applyCodeIntel();
-  }, 200);
+  }, 100);
 
   useEffect(() => {
     debouncedUpdate();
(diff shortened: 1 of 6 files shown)


src/components/molecules/Monaco/useCodeIntel.ts
```diff
@@ -34,31 +34,32 @@ function useCodeIntel(
...
+          disposablesRef.current = newDisposables;
+        });
     }
   };
 
-  useDebounce(
-    () => {
-      clearCodeIntel();
-      applyCodeIntel();
+  useEffect(() => {
```
olensmar, line 55: @devcatalin I had to go back to useEffect since useDebounce wasn't calling the cleanup function - perhaps there is some way to combine these two!? to still achieve a debounce on property changes? (since many change quite often)..

devcatalin: we can keep the `useEffect` hook and use the `debounce` function from lodash instead

olensmar: thanks I'll try that

olensmar: ok - debounce added - hope I got it right... thanks @devcatalin

devcatalin (approved):
lgtm!

PR #807, opened 2021-11-30
feat: improved New Resource wizard

Toxer10:
## Changes

- improved New Resource wizard


package.json
```diff
@@ -103,6 +104,7 @@
     "micromatch": "4.0.4",
     "module-alias": "2.2.2",
     "monaco-yaml": "3.2.1",
+    "mousetrap": "^1.6.5",
```
devcatalin, line 107: we are using `react-hotkeys-hook` in the `HotKeysHandler` component, could you try using that for the shortcuts instead of adding `mousetrap` ?

devcatalin (changes requested):

package.json
Toxer10: Sure

devcatalin (approved):

olensmar (approved):
nice!

PR #822, opened 2021-12-02
feat: add create new resource option for folder item

erdkse:
This PR...

## Changes

- Adds a new option to file items in treeview to create a new resource

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/FileTreePane/FileTreePane.tsx
+++ b/src/components/organisms/FileTreePane/FileTreePane.tsx
@@ -25,6 +25,7 @@ import {selectFile, setSelectingFile} from '@redux/reducers/main';
 import {
   closeFolderExplorer,
   openCreateFolderModal,
+  openNewResourceWizard,
   openRenameEntityModal,
   setShouldExpandAllNodes,
 } from '@redux/reducers/ui';
@@ -321,6 +322,7 @@ interface TreeItemProps {
   onExcludeFromProcessing: (relativePath: string) => void;
   onIncludeToProcessing: (relativePath: string) => void;
   onCreateFolder: (absolutePath: string) => void;
+  onCreateResource: (targetFolder: string) => void;
   isExcluded?: Boolean;
   isFolder?: Boolean;
 }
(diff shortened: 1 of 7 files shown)


src/components/organisms/FileTreePane/FileTreePane.tsx
```diff
@@ -372,6 +375,8 @@ const TreeItem: React.FC<TreeItemProps> = props => {
     ? fileMap[ROOT_FILE_ENTRY].filePath
     : path.join(fileMap[ROOT_FILE_ENTRY].filePath, treeKey);
 
+  const targetFolder = isRoot ? ROOT_FILE_ENTRY : treeKey.replace(path.sep, '');
```
olensmar, line 378: hm.. I think targetFolder should be relative to root - or empty for root!? Does this work for nested folders?

erdkse: It creates files in the nested folders but can't be expanded. (might not be related)

I followed the structure of the folder dropdown in NewResourceWizard.

olensmar:
looks good although I noticed that new resource files are not shown in the file-explorer - maybe they don't get added to the filemap!? or the fileexplorer doesn't update on changes?

erdkse:
I fixed that - actually there is a seamless way to solve it and tried for that but gave up because lack of time - will continue working on it

src/models/ui.ts
```diff
@@ -95,6 +95,7 @@ export type UiState = {
   resetLayout: boolean;
   clusterPaneIconHighlighted: boolean;
   clusterStatusHidden: boolean;
+  reloadFileTree: boolean;
```
olensmar, line 98: hm... why is this necessary? doesn't the file-tree react to changes in the fileMap ?

erdkse: It does - but there is another problem I added this as temporary

olensmar (approved):

olensmar:
nice work @erdkse !

PR #832, opened 2021-12-03
fix: fixed behavior when the user could add a new namespace to the list of namespaces

Toxer10:
## Fixes

- fixed behavior when the user could add a new namespace to the list of namespaces


olensmar (approved):
thank you!

Issue #852, opened 2021-12-06
Incoming links from deleted resource still showing


olensmar:
Reproduce:
- find resource A with an incoming link from another resource B
- delete the other resource B
- the incoming links popup for resource A is still showing the incoming link from resource B - expected that this would have been removed

PR #860, opened 2021-12-08
fix: removed all referring refs when deleting a resource

olensmar:
A side-benefit of this is that we now have an isOptional flag on refs - which we could use in the ref popups (and anywhere else) to indicate that a ref is optional (goes for both satisfied and unsatisfied refs)

## Changes

-

## Fixes

- #852 

## How to test it

-

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/redux/reducers/main.ts
+++ b/src/redux/reducers/main.ts
@@ -25,7 +25,7 @@ import {
 } from '@models/appstate';
 import {K8sResource} from '@models/k8sresource';
 
-import {findResourcesToReprocess} from '@redux/services/resourceRefs';
+import {findResourcesToReprocess, removeReferringRefs} from '@redux/services/resourceRefs';
 import {resetSelectionHistory} from '@redux/services/selectionHistory';
 import {performResourceDiff} from '@redux/thunks/diffResource';
 import {loadClusterDiff} from '@redux/thunks/loadClusterDiff';
@@ -357,6 +357,9 @@ export const mainSlice = createSlice({
       if (!resource) {
         return;
       }
+
+      removeReferringRefs(resource, state.resourceMap);
+
       if (state.selectedResourceId === resourceId) {
         clearResourceSelections(state.resourceMap);
         state.selectedResourceId = undefined;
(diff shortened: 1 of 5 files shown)


src/redux/reducers/main.ts
```diff
@@ -357,6 +357,9 @@ export const mainSlice = createSlice({
       if (!resource) {
         return;
       }
+
+      removeReferringRefs(resource, state.resourceMap);
```
olensmar, line 361: @devcatalin @erdkse not sure we need to call this method from elsewhere in other scenarios/actions!?

src/redux/services/fileEntry.ts
```diff
@@ -530,17 +531,8 @@ export function removePath(absolutePath: string, state: AppState, fileEntry: Fil
     }
   }
 
-  reprocessResources(
```
olensmar, line 533: @devcatalin @erdkse removed this since I couldn't think of anything that needs to be reprocessed except for refs that are removed by the line below instead

src/redux/services/resourceRefs.ts
```diff
@@ -215,6 +215,27 @@ function cleanResourceRefs(resources: K8sResource[]) {
   });
 }
 
+/**
+ * Removes all refs to the specified resource from other resources
+ */
+
+export function removeReferringRefs(resource: K8sResource, resourceMap: ResourceMapType) {
```
devcatalin, line 222: shouldn't we transform the refs to unsatisfied instead of removing them? - and of course keep in mind the "optional" sibling

src/redux/reducers/main.ts
devcatalin: this looks like the only place

src/redux/services/resourceRefs.ts
olensmar: yes - nice catch! will fix...

devcatalin (approved):
LGTM!

olensmar:
@devcatalin i've update this to also handle optional refs correctly - i.e. if a ref was option it gets removed on delete instead of changed to unsatisfied. A side-effect of this is that we now have an isOptional flag on refs - which we could use in the ref popups (and anywhere else) to indicate that a ref is optional

PR #870, opened 2021-12-10
feat: add new resources to existing file

erdkse:
This PR...

## Changes

- adds option to save new resource to existing file

## Fixes

- Change style of load/reload action button

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/FileTreePane/FileTreePane.tsx
+++ b/src/components/organisms/FileTreePane/FileTreePane.tsx
@@ -322,7 +322,7 @@ interface TreeItemProps {
   onExcludeFromProcessing: (relativePath: string) => void;
   onIncludeToProcessing: (relativePath: string) => void;
   onCreateFolder: (absolutePath: string) => void;
-  onCreateResource: (targetFolder: string) => void;
+  onCreateResource: (params: {targetFolder?: string; targetFile?: string}) => void;
   isExcluded?: Boolean;
   isFolder?: Boolean;
 }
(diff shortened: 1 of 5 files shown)


src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -334,21 +365,29 @@ const NewResourceWizard = () => {
...
-        {fileMap[ROOT_FILE_ENTRY] ? (
-          <SaveToFolderWrapper>
-            <StyledCheckbox onChange={e => setShouldSaveState(e.target.checked)} checked={shouldSaveToFolder}>
-              Save to folder
-            </StyledCheckbox>
+
+        <SaveDestinationWrapper compact>
+          <StyledSelect value={savingDestination} onChange={handleSavingDestinationChange}>
+            <Option value="saveToFolder">Save to folder</Option>
+            <Option value="saveToFile">Save to file</Option>
```
olensmar, line 372: let's call this "Add to existing file"

olensmar:
need to fix sizing - 
image

src/components/organisms/FileTreePane/FileTreePane.tsx
```diff
@@ -425,6 +414,17 @@ const TreeItem: React.FC<TreeItemProps> = props => {
         onClick={e => {
           e.domEvent.stopPropagation();
 
+          onCreateResource(isFolder ? {targetFolder: target} : {targetFile: target});
+        }}
+        key="create_resource"
+      >
+        New Resource
```
olensmar, line 421: Can we call it "Add Resource" for files and "New Resource" for folders?

erdkse: yes

olensmar:
thanks @erdi - see below - could we make the left dropdown smaller so more of the filename is shown? perhaps we should call the option "Add to file"  to make it even smaller?

image

olensmar (approved):
great - thank you!

erdkse:
> thanks @erdi - see below - could we make the left dropdown smaller so more of the filename is shown? perhaps we should call the option "Add to file" to make it even smaller?
> 
> image

I see this problem but couldn't produce a better idea to solve - implemented your suggestion - there will be always wider file names but users can read whole file name when they hover the selection - I thought that kind a solution and left sizes as it was

annnd you mentioned someone else 😄

PR #882, opened 2021-12-13
feat: Cluster Compare modal resizable

topliceanurazvan:
## Changes

- Resizable Cluster Compare Modal
- Resizable Resource Diff Modal

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/ResourceDiff/ResourceDiff.tsx
+++ b/src/components/molecules/ResourceDiff/ResourceDiff.tsx
@@ -1,5 +1,6 @@
 import {useMemo, useState} from 'react';
 import {MonacoDiffEditor} from 'react-monaco-editor';
+import {useMeasure} from 'react-use';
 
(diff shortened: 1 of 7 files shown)


devcatalin:
Need to fix two things:
- the Skeleton is out of the modal body when Cluster Compare initially loads
- ItemRenderer styling because the Suffix icons were moved to far right in the Navigator

olensmar:
thanks @topliceanurazvan - looking good! It seems this doesn't apply to the standalone Diff - will that be in this PR also or in a separate one?

topliceanurazvan:
> thanks @topliceanurazvan - looking good! It seems this doesn't apply to the standalone Diff - will that be in this PR also or in a separate one?

@olensmar We will add it in this PR.

devcatalin (approved):
Nice work!

PR #889, opened 2021-12-14
ref parsing/processing performance

olensmar:
improved parsing and ref-processing performance substantially - for around 1000 resources time went from 1200 ms to 20ms


devcatalin (approved):
Nice work, @olensmar 🙌 All the improvements look great!

olensmar:
thanks @devcatalin - it's a bit scary to change these things - we'll have to keep our eyes open for weird behaviour...

PR #901, opened 2021-12-17
feat: apply resource/resources to namespace

topliceanurazvan:
## Changes

- Use modal confirm component for applying resource/resources
- Apply resource/resources to selected namespace 
- Option to create a new namespace if it does not exist in the cluster and apply the resource/resources to it

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
@topliceanurazvan looking good Razvan - i'm wondering if we could populate the list of namespaces with the namespaces defined in the target cluster!? You could use the NamespaceHandler.listResourcesInCluster to get this list (with some post processing required) . 

Also - perhaps we should have give users the possibility to create a new namespace in this dialog - i.e. if the user specifies a namespace that doesn't exist (in the list of found namespaces) we prompt if they want to create the namespace instead - we could use the CoreV1Api.createNamespace of the kubernetes-client method for this!? Or should that be a separate issue?

what do you think @erdkse @devcatalin ?

src/components/molecules/CheckedResourcesActionsMenu/CheckedResourcesActionsMenu.tsx
```diff
@@ -68,6 +70,18 @@ const CheckedResourcesActionsMenu: React.FC = () => {
   const isInPreviewMode = useAppSelector(isInPreviewModeSelector);
   const resourceMap = useAppSelector(state => state.main.resourceMap);
 
+  const [isApplyModalVisible, setIsApplyModalVisible] = useState(false);
+
+  const checkedResources = useMemo(
+    () => checkedResourceIds.map(resource => resourceMap[resource]).filter((r): r is K8sResource => r !== undefined),
+    [checkedResourceIds]
```
devcatalin, line 77: don't forget to add `resourceMap` to the dependency array here

topliceanurazvan:
@devcatalin @olensmar I've added the things discussed above and the PR is ready to be reviewed.

olensmar:
looks good - but unfortunately i'm not seeing all namespaces from my cluster (see list below):

image

image

src/components/molecules/CheckedResourcesActionsMenu/CheckedResourcesActionsMenu.tsx
```diff
@@ -68,6 +70,18 @@ const CheckedResourcesActionsMenu: React.FC = () => {
...
 
+  const [isApplyModalVisible, setIsApplyModalVisible] = useState(false);
+
+  const checkedResources = useMemo(
+    () => checkedResourceIds.map(resource => resourceMap[resource]).filter((r): r is K8sResource => r !== undefined),
+    [checkedResourceIds, resourceMap]
+  );
+
+  const confirmModalTitle = useMemo(
+    () => `Deploy selected resources (${checkedResourceIds.length}) to cluster [${currentContext || ''}]?`,
```
devcatalin, line 81: let's use the length of `checkedResources` instead of `checkedResourceIds` so we can see if we get different lengths, although it shouldn't happen

src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
```diff
@@ -0,0 +1,191 @@
...
+  margin-top: 16px;
+`;
+
+const TitleContainer = styled.div`
+  display: flex;
+  align-items: center;
+`;
+
+interface IProps {
+  isModalVisible: boolean;
```
devcatalin, line 42: we could rename this to just `isVisible`

src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
```diff
@@ -0,0 +1,191 @@
...
+          if (err.statusCode === 409) {
+            setErrorMessage('Namespace already exists in the cluster!');
+          } else {
+            setErrorMessage(err.message);
+          }
+        });
+    } else if (selectedOption === 'existing') {
+      onOk(selectedNamespace);
+    }
+  }, [createNamespaceName, selectedNamespace, selectedOption]);
```
devcatalin, line 98: this has a few dependencies missing, check the warning below

src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
```diff
@@ -0,0 +1,191 @@
...
+    if (!namespaces.includes(defaultNamespace)) {
+      setSelectedOption('create');
+      setSelectedNamespace('default');
+      setCreateNamespaceName(defaultNamespace);
+    } else {
+      setSelectedOption('existing');
+      setSelectedNamespace(defaultNamespace);
+      setCreateNamespaceName('');
+    }
+  }, [namespaces]);
```
devcatalin, line 110: see the warning below

src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
```diff
@@ -0,0 +1,191 @@
...
+          <NamespaceContainer>
+            <span>Namespace:</span>
+            <Select
+              value={selectedNamespace}
+              showSearch
+              defaultValue={defaultNamespace}
+              onChange={value => setSelectedNamespace(value)}
+            >
+              {namespaces.map(namespace => {
+                if (typeof namespace !== 'string') {
```
devcatalin, line 153: use `.filter` instead: `namespaces.filter(n => typeof n !== 'string').map...`

src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
```diff
@@ -0,0 +1,191 @@
...
+              <Input
+                autoFocus
+                defaultValue={createNamespaceName}
+                placeholder="Enter namespace name"
+                value={createNamespaceName}
+                onChange={e => {
+                  setCreateNamespaceName(e.target.value);
+
+                  if (errorMessage) {
+                    setErrorMessage('');
```
devcatalin, line 178: should this be `setErrorMessage(errorMessage)` ?

src/components/organisms/ActionsPane/ActionsPane.tsx
```diff
@@ -90,8 +90,11 @@ const ActionsPane = (props: {contentHeight: string}) => {
   const isClusterDiffVisible = useAppSelector(state => state.ui.isClusterDiffVisible);
   const isActionsPaneFooterExpanded = useAppSelector(state => state.ui.isActionsPaneFooterExpanded);
   const kubeconfigPath = useAppSelector(state => state.config.kubeconfigPath);
-  const [key, setKey] = useState('source');
+
+  const [isApplyModalVisible, setIsApplyModalVisible] = useState(false);
   const [isButtonShrinked, setButtonShrinkedState] = useState<boolean>(true);
+  const [key, setKey] = useState('source');
```
devcatalin, line 96: what "key" is this? could we find a better variable name? - although I don't think this is from your PR but I just noticed it now

src/components/molecules/ResourceDiff/ResourceDiff.tsx
```diff
@@ -105,6 +110,14 @@ const ResourceDiff = (props: {
     readOnly: true,
   };
 
+  const confirmModalTitle = useMemo(
```
devcatalin, line 113: I saw that the `confirmModalTitle` is repeated across multiple components, shouldn't we move it to the Modal instead?

edit: I understood why it's repeated, I'm wondering if we should make a custom hook for this?

src/hooks/useTargetClusterNamespaces.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  const [namespaces, setNamespaces] = useState<string[]>([]);
+
+  useEffect(() => {
+    const setClusterNamespaces = async () => {
+      const kc = new k8s.KubeConfig();
+      kc.loadFromFile(kubeconfigPath);
+      kc.setCurrentContext(context || '');
+
+      const ns = await NamespaceHandler.listResourcesInCluster(kc);
```
devcatalin: let's move this and the 3 lines above into the `getTargetClusterNamespaces` function

src/utils/resources.ts
```diff
@@ -102,3 +102,22 @@ export function diffLocalToClusterResources(localResource: K8sResource, clusterR
     cleanClusterResourceContent,
   };
 }
+
+export function getDefaultNamespace(resources: K8sResource[]) {
```
devcatalin, line 106: should this be called `getDefaultNamespaceForApply` ?

src/utils/resources.ts
```diff
@@ -102,3 +102,22 @@ export function diffLocalToClusterResources(localResource: K8sResource, clusterR
...
+  let namespace = 'default';
+
+  resources.forEach(resource => {
+    if (resource.namespace) {
```
devcatalin, line 110: you could remove a level of nesting here by doing:
```
if (!resource.namespace) {
    return;
}
```

devcatalin (changes requested):

topliceanurazvan:
@olensmar I've committed some changes and updates. Could you please test it again and see if now there are shown all the namespaces in the dropdown?

src/constants/makeApplyText.ts
```diff
@@ -0,0 +1,7 @@
+export const makeApplyKustomizationText = (name: string, context: string) =>
```
devcatalin, line 1: Let's chabge the context type to `string | undefined` and use it as `${context || ''}` below so we don't have to do it where the function is being used

olensmar:
> @olensmar I've committed some changes and updates. Could you please test it again and see if now there are shown all the namespaces in the dropdown?

Thanks @topliceanurazvan - I get all my namespaces now: 

image

src/redux/thunks/applyResource.ts
```diff
@@ -36,12 +36,13 @@ function applyKustomization(
   fileMap: FileMapType,
   kubeconfig: string,
   context: string,
+  namespace: string,
```
olensmar, line 39: shouldn't this be optional? I'm thinking that in case the kustomization generates resources for multiple namespaces the won't want to specify this at all - since it would force all resources into the same namespace..

olensmar:
@topliceanurazvan i'm missing this when deploying a helm-chart - both `helm install` and `helm template` commands have the `-n namespace` argument for setting the target namespace, and a `--create-namespace` argument to create the specified namespace is missing (which we could use instead of creating the namespace ourselves) - see https://helm.sh/docs/helm/helm_install/ and https://helm.sh/docs/helm/helm_template/

src/redux/thunks/applyResource.ts
topliceanurazvan: @olensmar Should I add a third option in the modal ( `Use kustomization namespaces` ) or should we not show the namespace selector at all?

olensmar: perhaps there could be a "<as configured>" option which is selected by default and would result in Monokle not setting the namespace explicitely? (this would apply to all deploy actions)

devcatalin: @olensmar wouldn't that be confusing from the UX perspective? The users might think that a namespace is required for deployment.. I think a third radio button would be better, but we need to decide on the label

devcatalin: so we have "Use existing namespace", "Create new namespace" and the third one could be "None" ? or "Neither" ?

olensmar: sorry - my bad - "Use Existing" is fine as long as it doesn't override what is in the manifests

devcatalin (changes requested):

src/components/molecules/HelmChartModalConfirmWithNamespaceSelect/HelmChartModalConfirmWithNamespaceSelect.tsx
```diff
@@ -0,0 +1,145 @@
...
+        return;
+      }
+
+      onOk(createNamespaceName, true);
+    } else if (selectedOption === 'existing') {
+      onOk(selectedNamespace);
+    } else if (selectedOption === 'none') {
+      onOk();
+    }
+  }, [createNamespaceName, selectedOption]);
```
devcatalin, line 70: Missing 'onOk' and 'selectedNamespace' from dependencies

devcatalin (approved):
Nice work!

olensmar (approved):

Issue #841, opened 2021-12-03
Action to save cluster resource locally


olensmar:
I would like to save resources in my cluster or during a preview to a local file via a dedicated "Save to file.." action in the resource actions popup

olensmar:
cc @kylehodgetts

olensmar:
this could be part of #730

devcatalin:
For the similar action in Cluster Compare:

We decided to show details about what actions we'll do to each resource in the confirm modal:
- list the resources that will be replaced
- list the resources that will be saved + options to save to a folder, or a file, or leave them as unsaved resources

@tommywhitehead

tommywhitehead:
@devcatalin can I see what you have so far?

mortada-codes:
@jfermi 



https://user-images.githubusercontent.com/20525304/150302615-dbd7bf28-5c9a-4eae-b80a-1ca20f6a1571.mp4

PR #909, opened 2021-12-20
feat: option to show cluster only resources in cluster compare

devcatalin:
This PR...

## Changes

- added "Show all" option under "Cluster resources" title
- improved Cluster Compare styling

## Fixes

-

## How to test it

-

## Screenshots

image

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan (changes requested):
Nice styling refactor! 👍🏻 

I found a problem though, somehow the width of the modal can get lower than the one of the grid container, and it hides the scrollbar and some content from the right column ( see the attached screenshot ).

image

olensmar:
thanks @devcatalin - looks good! 

I found a couple of issues - not sure they are related - see screenshot below:
- when selecting resources that exist only in the cluster and not locally Monokle enables the "deploy selected resources to cluster" option - I think it should enable the other button instead!?
- the count displayed in the enabled button is for all resources - not the just the filtered/selected ones

image

devcatalin:
I pushed two commits fixing the following:
- styling/responsiveness issue
- clearing selected matches when filters are changed
- disabled/enabled state of the deploy and replace buttons

@olensmar , I have the following questions:
1. Should we disable changing filters when a selection is made or is it okay to clear selections?
2. The replace is currently enabled only when both local resource and cluster resource exist. 
To enable it when only the cluster resource exist, we have to decide how we would save the resources.
Should I just create unsaved resources? 
Or do we add an inputs to select a folder or a file where the resources should be saved? - in this case maybe the word "Replace" is not that great

devcatalin:
For the second question related to "replace" action, I could do it in a separate PR when I implement this: https://github.com/kubeshop/monokle/issues/841

devcatalin:
@olensmar @topliceanurazvan this is ready for review & merge

topliceanurazvan (changes requested):
Great work! 👍🏻 The only problem that I encountered is that with some filters applied ( the namespace in my case), after clicking on the `Select all` checkbox, that number of selected resources is much bigger than the actual selected resources ( see the attached screenshot).

image

devcatalin:
fixed the "select all" checkbox

topliceanurazvan (approved):

PR #925, opened 2021-12-23
fix: height offset of panes in preview mode

devcatalin:
This PR...

## Changes

- this is a temporary fix for calculating the heights of panes while in preview mode
- we should refactor the PaneManager implementation so we can get the values dinamically

## Fixes

- either the footer or part of the header were hidden during preview mode

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan (approved):
lgtm!

PR #927, opened 2021-12-23
Improve ESLint and TS configurations

devcatalin:
This PR...

## Changes

- better ordering/organization of rules in estlintrc
- added multiple typescript lint rules, set all of them to "warn" to begin with
- set `noUncheckedIndexedAccess` to true in tsconfig, read more: https://devblogs.microsoft.com/typescript/announcing-typescript-4-1-beta/#no-unchecked-indexed-access

## Fixes

- fixed all errors introduced by `noUncheckedIndexedAccess`

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -9,3 +9,8 @@ server.js
 build.js
 init.js
 site
+commitlint.config.js
+craco.config.js
+notarization
+resources
+.eslintrc.js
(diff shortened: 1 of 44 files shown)


topliceanurazvan (approved):

devcatalin:
I decided to close this because adding `noUncheckedIndexedAccess` at this point makes it harder for us to work with `resourceMap`, `fileMap` and similar structures. We might get some improvements in the 4.6 version of TypeScript so we could decide then if we want to enable this option.

Nevertheless, I will make a separate PR that will only add ESLint rules set to warning.

PR #931, opened 2021-12-27
feat: api version dropdown + group kinds by api

topliceanurazvan:
This PR...

## Changes

- Groups resource kind dropdown options by api version
- Add Select with search for Api versions
- Filter resource kind on api version select

## Fixes

- Reset state when closing the wizard
- Set corresponding `Resource kind` dropdown options and `filteredResources` when cloning a resource

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan:
@olensmar @tommywhitehead - I have a question, right now if a `Resource Kind` is already selected and you select an `API Version` which doesn't contain the currently selected `Resource kind`, we automatically change the `Resource Kind` with the first option from the list of corresponding `Kinds` of the newly selected API Version. Is it okay this way or should we change it to `undefined` instead?

olensmar:
thanks @topliceanurazvan - I think this is fine although I wonder if we could limit the list of versions to contain only those versions that are available for the selected resource-kind (if a resource-kind is selected...) ? 

I'm not even sure we maintain a list of available versions - perhaps this should be exposed by the corresponding ResourceKindHandler? @devcatalin what do you think?

topliceanurazvan:
@olensmar - Also by limiting the list of versions when a resource-kind is selected, maybe we should think of a way to reset/deselect the api-version/resource-kind selection, otherwise the user will be limited to choose only from the api-versions corresponding to the selected resource-kind and vice-versa.

devcatalin:
@olensmar right now we only specify one version for each kind using the `clusterApiVersion` from kind handlers, so we would be limiting the versions to only one option.
@topliceanurazvan has a good point, if we limit the options two-way then there should be some way to reset them..

I suggest we don't limit the api version options, but set the kind to undefined if the api version is changed

topliceanurazvan:
@tommywhitehead Any suggestions on how we should set the `Resource kind` when changing the api version?

topliceanurazvan:
@olensmar @devcatalin The PR is ready to be reviewed.

devcatalin (approved):
LGTM, nice work!

PR #940, opened 2021-12-29
dynamic kindhandlers

olensmar:
This PR introduces the dynamic discovery and installation of kindhandlers based on CRDs in resources folder and in found CustomResourceDefinitions at runtime. It also adds CRDs for 
- argocd
- argo-rollouts
- emissary
- testkube

it's mostly crd/handler files added/removed - not so much code...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan (approved):
lgtm!

olensmar:
@devcatalin would love your input on this before merging...

src/kindhandlers/common/customObjectKindHandler.ts
```diff
@@ -21,19 +32,143 @@ function extractResourceVersion(resource: K8sResource, kindVersion: string, kind
...
+    delete schema.properties['apiVersion'];
+    delete schema.properties['kind'];
+    delete schema.properties['metadata'];
+  }
+
+  return schema;
+}
+
+export function extractKindHandler(crd: any, handlerPath?: string) {
+  const spec = crd.spec;
```
devcatalin, line 48: Should we check if crd is not undefined here, and the same for spec?

src/kindhandlers/common/formLoader.ts
```diff
@@ -6,7 +6,10 @@ const uiformSchemaCache = new Map<string, any>();
 export function getFormSchema(kind: string) {
   try {
     if (!formSchemaCache.has(kind)) {
-      formSchemaCache.set(kind, JSON.parse(loadResource(`form-schemas/${kind.toLowerCase()}-schema.json`)));
+      const formSchema = loadResource(`form-schemas/${kind.toLowerCase()}-schema.json`);
```
devcatalin, line 9: Use `path.sep` here instead of `/`

src/kindhandlers/common/formLoader.ts
```diff
@@ -18,7 +21,10 @@ export function getFormSchema(kind: string) {
 export function getUiSchema(kind: string) {
   try {
     if (!uiformSchemaCache.has(kind)) {
-      uiformSchemaCache.set(kind, JSON.parse(loadResource(`form-schemas/${kind.toLowerCase()}-ui-schema.json`)));
+      const uiSchema = loadResource(`form-schemas/${kind.toLowerCase()}-ui-schema.json`);
```
devcatalin, line 24: same as above, should use `path.sep`

src/kindhandlers/index.ts
```diff
@@ -99,6 +82,22 @@ const HandlerByResourceKind = Object.fromEntries(
   }).map(kindHandler => [kindHandler.kind, kindHandler])
 );
 
+export function registerKindHandler(kindHandler: ResourceKindHandler, replace: boolean) {
```
devcatalin, line 85: can we rename `replace` to `shouldReplace` ?

src/kindhandlers/index.ts
```diff
@@ -139,10 +138,40 @@ export const getDependentResourceKinds = (resourceKinds: string[]) => {
...
+        });
+      }
+    } catch (e) {
+      log.warn(`Failed to parse kindhandler CRD at ${crdPath}`, e);
+    }
+  });
+}
 
-  return refMapper.target.kind === kind;
+function findFiles(dir: string, ext: string) {
```
devcatalin, line 164: Could we rewrite this to be async and use the promise version of `fs` methods?
It would be nice to start doing this for new features so we won't have to refactor more code later if we'll decide to rewrite all I/O functions to be async

src/models/resourcekindhandler.ts
```diff
@@ -30,19 +23,22 @@ interface RefMapper {
...
   };
   target: {
     kind: string;
     pathParts?: string[];
   };
 
   type: 'path' | 'name' | 'pairs';
+
+  // called to validate if an unsatisfied ref should be created for this mapper
+  validateUnsatisfiedRef?: (
```
devcatalin, line 37: Would `shouldCreateUnsatisfiedRef` be a better name for this?

src/redux/services/resourceRefs.ts
```diff
@@ -68,6 +63,14 @@ const addRefNodeAtPath = (refNode: RefNode, path: string, refNodesByPath: Record
   }
 };
 
+export function refMapperMatchesKind(refMapper: RefMapper, kind: string) {
+  if (refMapper.target.kind.startsWith('$')) {
+    return kind.match(refMapper.target.kind.substring(1)) !== null;
```
devcatalin, line 68: I'm wondering if it would be safer/faster to split by `|` and check each value manually instead of regex, but we could leave it like this to start with

src/redux/services/resourceRefs.ts
```diff
@@ -359,13 +361,39 @@ function shouldCreateUnsatisfiedRef(
...
+
   return true;
 }
 
+function getSiblingValues(outgoingRefMapper: RefMapper, sourceResource: K8sResource, sourceRefNode: RefNode) {
+  const siblingValues: Record<string, string> = {};
+  if (outgoingRefMapper.source.siblingMatchers) {
+    Object.keys(outgoingRefMapper.source.siblingMatchers).forEach(key => {
+      const value = getSiblingValue(key, outgoingRefMapper, sourceResource, sourceRefNode);
+      if (typeof value === 'string') {
```
devcatalin, line 389: Shouldn't we include numbers and booleans, or are those handled?

src/redux/services/schema.ts
```diff
@@ -7,8 +7,11 @@ import {isKustomizationResource} from '@redux/services/kustomize';
 
 import {getResourceKindHandler} from '@src/kindhandlers';
 
+// @ts-ignore
 const k8sSchema = JSON.parse(loadResource('schemas/k8sschemas.json'));
```
devcatalin, line 11: Should we make a function that does `JSON.parse` and `loadResource` in a try catch and use that instead? This would crash the app if the parsing fails

src/redux/services/schema.ts
```diff
@@ -75,3 +83,35 @@ export function loadCustomSchema(schemaPath: string, resourceKind: string): any
     schemaCache.set(resourceKind, undefined);
   }
 }
+
+export function extractSchema(crd: any, versionName: string) {
+  const versions: any[] = crd.spec.versions;
```
devcatalin, line 88: This should be `const versions: any[] = crd?.spec?.versions || [];`

src/redux/services/schema.ts
```diff
@@ -75,3 +83,35 @@ export function loadCustomSchema(schemaPath: string, resourceKind: string): any
...
+  const version = versions.find((v: any) => v.name === versionName);
+  const schema = version?.schema?.openAPIV3Schema ? version.schema.openAPIV3Schema : undefined;
```
devcatalin, line 90: This looks redundant, you can simply do `const schema = version?.schema?.openAPIV3Schema;`

src/redux/services/schema.ts
```diff
@@ -75,3 +83,35 @@ export function loadCustomSchema(schemaPath: string, resourceKind: string): any
...
+  if (schema) {
```
devcatalin, line 91: Personal preference, I would do `if (!schema) { return; }` here

devcatalin (changes requested):

src/redux/services/schema.ts
olensmar: in this specific case it was intentional to let the app crash if the this parsing fails - as it is required to run correctly - but perhaps we can handle the situation more gracefully..

devcatalin:
I added a few questions/changes requests but overall this looks awesome! Great job, Ole!
This PR will make Monokle much more powerful 😄

src/redux/services/resourceRefs.ts
olensmar: yes let's leave it for now - we discussed adding the apiVersion to this also eventually - which we could cover with regex expressions also

src/redux/services/schema.ts
devcatalin: Right, then we could leave it like this or we could show an error alert saying that some of the features failed to load. But I'm not sure if we assume that this works/exists in other places which might crash the app later if we don't crash it to start with, hmm..

devcatalin (approved):

src/kindhandlers/index.ts
olensmar: done - please have a look !

olensmar:
@devcatalin after making the loading of custom kindhandlers asynchronous I had to add a delay in the test that depends on them - since it was occasionally run before the async loading had finished... perhaps there is a better way to "wait" for these kindhandlers to load!?

devcatalin:
@olensmar you could try doing this `await new Promise(process.nextTick);` or `await new Promise(setImmediate);` at the start of the test, but I'm not 100% sure that it will work as intended.

Another solution would be to implement something like this in the file/module where the kindHandlers are loaded:

```typescript
import {EventEmitter} from 'events';

const KindHandlersEventEmitter = new EventEmitter();

// use this to trigger the "loaded" event when the async loading of kind handlers has finished
KindHandlersEventEmitter.emit('loaded');

// this can be awaited 
// alternatively, we can wrap this promise in the 'promiseTimeout' helper from @utils/promises if we want this to be rejected after a certain amount of time
export const waitKindHandlersLoading = new Promise<void>(resolve => {
  KindHandlersEventEmitter.once('loaded', () => {
    resolve();
  });
});
```

olensmar:
thanks @devcatalin - i like the event-emitter approach - i'll give that a try!

olensmar:
thanks @devcatalin - that seems to have worked... have a look https://github.com/kubeshop/monokle/pull/940/files#diff-9234b3cdbf49061b1e598aec9779e0ff80145398639d4141caf44cd9fde0b84eR181

Issue #332, opened 2021-09-02
use monacos yaml-language-server for resource validation


olensmar:
to provide consistent validation results / error messages

PR #960, opened 2022-01-07
feat: improved syntax error messages and position, added links to popup

olensmar:
## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [X] added a test

image


olensmar:
I think this also removes the need for #332 for now since the validation messages we get from ajv are the same as shown in the editor - i'm guessing they use ajv under the hood also

olensmar:
@tommywhitehead please let me know if you have any comments on the formatting in the popup

olensmar:
> Should we use dots instead of slashes for the object paths?

the path comes from the ajv validator as a JsonPointer - we can replace the slashes with dots if that makes it easier to read!?

devcatalin (approved):
LGTM

devcatalin:
> the path comes from the ajv validator as a JsonPointer - we can replace the slashes with dots if that makes it easier to read!?

Maybe another question for Thomas?

PR #961, opened 2022-01-07
feat : hotkey navigation popup

topliceanurazvan:
This PR creates a quick search popup for applying `kind` or `namespace` filter, or select a resource.

## Changes

- `CTRL/CMD Shift P` hotkey to open the quick search
- Updated docs with missing hotkeys

## Fixes

- `ScrollIntoView` not getting the element into view in some cases.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test


olensmar:
this is great @topliceanurazvan - i'm wondering if in the situation below we could have some way to set "argo" as the name filter in the filter panel!? perhaps a "Filter all" link/button next to the "Resources" label!? @tommywhitehead could that work?

image

devcatalin:
@olensmar I think it should be a separate option in the dropdown so it can be selected with arrow keys.

Maybe we should have a group named "Actions" where we add that option, or it could be the first option under Resources..

olensmar:
good idea @devcatalin - perhaps we do this in a separate PR so we get this one merged?

olensmar (commented):
lgtm - the only thing I've noticed is that when pressing arrow-up to get to the last entry in the filtered list, and then pressing arrow-down to get back to the first entry, the label above the first entry is scrolled out of view: 

image

pressing arrow-up + arrow-down results in 

image

src/components/molecules/ScrollIntoView/ScrollIntoView.tsx
```diff
@@ -3,19 +3,35 @@ import React from 'react';
...
 };
 
-const ScrollIntoView = React.forwardRef(({children, id}: ScrollIntoViewProps, ref) => {
+const ScrollIntoView = React.forwardRef(({children, id, parentContainerElementId}: ScrollIntoViewProps, ref) => {
   const containerRef = React.useRef<HTMLLIElement>(null);
   React.useImperativeHandle(ref, () => {
     return {
       scrollIntoView: () => {
-        containerRef.current?.scrollIntoView({behavior: 'smooth', block: 'center', inline: 'nearest'});
+        // containerRef.current?.scrollIntoView({behavior: 'smooth', block: 'center', inline: 'nearest'});
```
DmitryEremenko, line 19: Do we need this commented out line? If not let's delete it

DmitryEremenko (changes requested):

olensmar (approved):
great!

DmitryEremenko (approved):

PR #962, opened 2022-01-07
fix: loading of editor schemas for bundled CRDs

olensmar:
This PR also fixes the New Resource dialog to contain all resource kinds

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
+++ b/src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
@@ -75,20 +75,16 @@ const NewResourceWizard = () => {
 
   const kindsByApiVersion = useMemo(
     () =>
-      ResourceKindHandlers.reduce((result, resource) => {
-        const foundResourceKindHandler = getResourceKindHandler(resource.kind);
-
-        if (foundResourceKindHandler) {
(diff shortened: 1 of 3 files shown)


src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -75,20 +75,18 @@ const NewResourceWizard = () => {
...
+          result[resourcekindHandler.clusterApiVersion].push(resourcekindHandler);
+        } else {
+          result[resourcekindHandler.clusterApiVersion] = [resourcekindHandler];
         }
 
         return result;
       }, {} as Record<string, ResourceKindHandler[]>),
-    []
+    // depend on resourceMap since newly loaded resources could have contained CRDs that resulted in dynamically
+    // created kindHandlers
```
olensmar, line 88: @devcatalin see comment in code - it would probably be better to have the kindhandlers be part of the app state so we can depend on those directly - is this what you are planning to solve the section blueprints?

PR #965, opened 2022-01-10
feat: multi resource actions

topliceanurazvan:
This PR adds `Delete` and `Save to file/folder` functionality for multiple selected resources.

## Changes

- Add `Save to file/folder` action for resources context menu in preview/cluster mode

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
thanks @topliceanurazvan - looks great! a couple of comments:
- suggest to make the "save to folder" option the default one
- if I select a bunch of resources, then add a filter that selects a subset of them, and then do the "Save.." action I'm still prompted to save the unfiltered list of objects - I expected to only export what is filtered (although I understand that more objects are actually selected).. would that work? or be confusing?
- what if I want to save to a folder "outside" my current project - could we provide an option for that? Perhaps "Save to project folder" shows the existing folder dropdown, "Export to folder" allows me to select any folder!?

@tommywhitehead thoughts?

topliceanurazvan:
@olensmar - Great catch with filtering when there are resources selected. I could think of some possible solutions, either we deselect all resources when we apply new filters ( with a confirm modal ), or as you said, do the actions only for the filtered resources that are selected, which might be a little bit confusing for the users. 

Right now we are disabling using the filters panel when there are selected resources and hiding the button for opening the panel. Maybe we should stick with not letting the users apply new filters when they have resources selected.

Any thoughts on this? @tommywhitehead @devcatalin

olensmar:
I'm getting the following when trying to download from a cluster without having opened a local folder - i.e. fileMap[ROOT_FILE_ENTRY] will be undefined

image

topliceanurazvan:
@olensmar - great catch! I've fixed it.

src/components/molecules/ChangeFiltersConfirmModal/ChangeFiltersConfirmModal.tsx
```diff
@@ -0,0 +1,34 @@
...
+import {extendResourceFilter, uncheckAllResourceIds} from '@redux/reducers/main';
+
+const ChangeFiltersConfirmModal: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const filtersToBeChanged = useAppSelector(state => state.main.filtersToBeChanged);
+
+  useEffect(() => {
+    if (filtersToBeChanged) {
+      Modal.confirm({
+        title: 'I understand that changing the filters will uncheck all my selected resources.',
```
olensmar, line 17: two thoughts:
- just use 'Changing the filter will uncheck all selected resources'
- can we uncheck only those that will be filtered away? - i.e. 'Changing the filter will uncheck all hidden resources'

topliceanurazvan: I think that would be a nice thing to have, but this would mean enabling back the Resource Filters Panel, which currently is disabled. And maybe we should think of another way of showing the panel, because the toggle button is hidden when you have selected resources ( maybe with some hotkey for opening it ).

olensmar: hm... why would we have to enable the filter panel in this scenario? The workflow I'm considering is:
- I select all deployments
- I realize I only want to export those in a specific namespace so I filter on that namespace by clicking on it in the navigator (or source editor)
- I now have to reselect all deployments (not a big deal...) - instead of keeping the selection on those deployments that are still visible

topliceanurazvan: I agree, but I think that we should be consistent, allowing the user to change the namespace also from the filters panel maybe? He will be able to do that through other actions, so he might get confused if that filter is disabled in the panel.

olensmar: I know we discussed this yesterday with @tommywhitehead also - just don't remember what we concluded - i guess i'm proposing that you can continue to filter your objects after selecting - just that whenever your filter hides a selected resource - that resource automatically gets unselected... !?

olensmar: @tommywhitehead thoughts?

topliceanurazvan: @olensmar @devcatalin - For now I unchecked the hidden resources when applying a filter outside the Filters Pane. But will leave the Filters Pane disabled, because it looks that it's a little bit more complex than expected to enable it when there are selected resources. We can make an issue for that if it's necessary and solve it in another PR. wdyt?

src/components/molecules/CheckedResourcesActionsMenu/CheckedResourcesActionsMenu.tsx
```diff
@@ -46,15 +54,23 @@ const StyledMenu = styled(Menu)`
...
   Modal.confirm({
     title,
     icon: <ExclamationCircleOutlined />,
     centered: true,
     onOk() {
+      let alertMessage = '';
       return new Promise(resolve => {
+        checkedResources.forEach(resource => {
+          dispatch(removeResource(resource.id));
+          alertMessage += `${alertMessage && ' | '}${resource.name}\n`;
```
devcatalin, line 69: shouldn't the template string be `${alertMessage} | ${resource.name}\n`?

src/components/molecules/SaveResourcesToFileFolderModal/SaveResourcesToFileFolderModal.tsx
```diff
@@ -0,0 +1,357 @@
...
+
+import {useFileExplorer} from '@hooks/useFileExplorer';
+
+import {removeIgnoredPathsFromResourceContent} from '@utils/resources';
+
+import Colors from '@styles/Colors';
+
+const {Option} = Select;
+
+const ErrorMessageLabel = styled.div`
```
devcatalin, line 33: let's move all styled components to a `styled.tsx` file and then
`import * as S from "./styled` to use them here

you can also remove "Styled" prefix from the component names in this case

devcatalin (changes requested):

src/components/molecules/CheckedResourcesActionsMenu/CheckedResourcesActionsMenu.tsx
topliceanurazvan: We want to the the delimiter ( `|` ) only if the message already contains a resource name.

devcatalin: okay, I'll close this

devcatalin (approved):
LGTM!

olensmar (approved):
good stuff!

PR #975, opened 2022-01-10
Update images in UI Overview

jfermi:
Updated images to 1.4.0 version of Monokle.  Added new settings and corrected grammar mistakes.

This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test


olensmar (approved):
lgtm!

docs/overview.md
```diff
@@ -1,43 +1,50 @@
...
-Monokle is layed out like many other desktop applications:
+Monokle is laid out like many other desktop applications:
 
-![Monokle UI](img/monokle-ui.png)
+![Monokle UI](img/monokle-ui-1.4.0.png)
 
 Left to right:
 
-- The vertical toolbar to the far left allows you to switch between File and Cluster mode.
+- The vertical toolbar to the far left allows you to switch between File, Kustomize, Helm and Cluster mode.
```
devcatalin, line 9: we should remove "and Cluster mode" from here since it was moved to the top header

olensmar: nice catch @devcatalin

jfermi:
I still see the cluster icon in the left toolbar under the Helm icon.  Am I
looking at an old version?

jfermi:
I see from this morning that 1.5.0 has this change.  I guess this begs the
question - for the work I am doing right now, should I go ahead and work
with 1.5.0 since the release is imminent?

olensmar:
yeah - that's what I was thinking also - perhaps we just make minor high-level changes to the existing 1.4.0 documentation and then double-down on the upcoming 1.5.0...

docs/overview.md
olensmar: or actually - this is for 1.4.0 - so it's ok for now... will have to change in the 1.5.0 documentation

PR #984, opened 2022-01-12
Fix element wrapping section headers

### LINKED ISSUE #958: Fix elements wrapping in section headers
image

From the picture above we can see that the expand button is moved to the next line. 
This happens when the navigator width is small.

We should probably strip characters from the section name and add ellipsis.

andreiv1992:
This PR...

## Changes

Added ellipsis to section headers

## Fixes

#958 
image 

## How to test it

Make the window really small

## Screenshots

image 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/SectionRenderer/SectionHeader.tsx
+++ b/src/components/molecules/SectionRenderer/SectionHeader.tsx
@@ -53,6 +53,23 @@ function SectionHeader(props: SectionHeaderProps) {
     return sectionInstance?.visibleDescendantItemIds?.length || 0;
   }, [sectionInstance.visibleDescendantItemIds]);
 
+  const { shouldDisplayPlus, shouldDisplayContext } = useMemo(() => {
+    let _shouldDisplayPlus = false;
+    let _shouldDisplayContext = false;
+    if (NameSuffix.Component) {
+      if (NameSuffix.options?.isVisibleOnHover === undefined) {
+        _shouldDisplayContext = true;
+      } else {
+        _shouldDisplayPlus = NameSuffix.options?.isVisibleOnHover && isHovered;
+      }
+    }
+
+    return {
+      shouldDisplayPlus: _shouldDisplayPlus,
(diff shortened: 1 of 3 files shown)


olensmar:
thanks @andreiv1992 - it looks like the cluster label has moved up to the right from below the "K8s Resources" label

image

andreiv1992:
Thanks a lot for catching that Ole.

Found a solution and it is working but I think it would be best if we can create a separate property(eg HeaderContext), similar to NameSuffix where we can display information like this on a new line. It is cumbersome to display information on a new line when it is in the middle of other inline components.

Let me know what you think

olensmar:
thanks @andreiv1992 - that sounds good to me - do you want to do that in this PR or start a separate branch after merging? @devcatalin what do you think?

src/components/molecules/SectionRenderer/SectionHeader.tsx
```diff
@@ -65,6 +65,15 @@ function SectionHeader(props: SectionHeaderProps) {
       dispatch(sectionInstance.checkable.uncheckItemsAction);
     }
   }, [sectionInstance, dispatch]);
+  let shouldDisplayPlus = false;
```
devcatalin, line 68: could you convert these to `useMemo` ?

devcatalin:
> thanks @andreiv1992 - that sounds good to me - do you want to do that in this PR or start a separate branch after merging? @devcatalin what do you think?

sounds good to me, as long as we don't break the UI somewhere

src/components/molecules/SectionRenderer/SectionHeader.tsx
andreiv1992: Sure

andreiv1992:
> 

I'd like to do it in a different PR

olensmar (approved):

PR #985, opened 2022-01-12
Update Resource Editing Doc for 1.4.0

jfermi:
Update images, correct grammar and update flow of document.  One minor change in Getting Started doc to correct misspelling.

This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test


docs/resource-editing.md
```diff
@@ -1,71 +1,78 @@
...
 
 ![Source Editor Broken Links](img/source-editor-broken-links.png)
-
+  
 ## Add Resource
 
 While using Monokle, you can directly add new K8s resources. Once you have browsed and added your project folder, click on the “New Resource” button at the top of the navigator to launch the “Add New Resource” dialog.
 
-![Add Resource](img/add-resource.png)
+![Add Resource](img/add-resource-1.5.0.png)
```
olensmar, line 37: should we have a 1.5.0 screenshot here?

jfermi: These screenshots will be that same for 1.4.0 and 1.5.0, so I renamed them 1.5.0 to note that these images will not need to be updated for the next release.  I will verify with the final release build to make sure, though.

jfermi: There are only a few of these.  I can come up with another process to keep up with this if it does not make sense.

olensmar: it makes sense, thanks for clarifying

olensmar (approved):

PR #1011, opened 2022-01-17
fix: set electron-builder version to 22.11.7

erdkse:
This PR...

attempts to fix pipeline errors

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "craco-less": "1.20.0",
     "cross-env": "7.0.3",
     "electron": "16.0.7",
-    "electron-builder": "22.11.11",
+    "electron-builder": "22.11.7",
     "electron-notarize": "1.1.1",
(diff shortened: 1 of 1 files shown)


olensmar:
what features/fixes are we "loosing" with this downgrade?

erdkse:
electron-builder documentation does not provide a really good sight but It looks like they do all releases together - such as electron-[builder|updater|publisher...]

These are all the things I can find, It didn't look me like it may break things, so I want to try - If fails we can revert it back;

https://github.com/electron-userland/electron-builder/releases/tag/v22.11.11
https://github.com/electron-userland/electron-builder/releases/tag/22.11.10
https://github.com/electron-userland/electron-builder/releases/tag/v22.11.9
https://github.com/electron-userland/electron-builder/releases/tag/v22.11.8

olensmar (approved):

PR #1023, opened 2022-01-18
fix: ignore parsing helm templates 

mortada-codes:
This PR...

## Changes

- added a check function to check if the resource is supported

## Fixes

- The app was crashing when opening a helm charts template with variables `{{ . }}` 

## How to test it

- open a helm charts project 
- should not see a helm chart templates

## Screenshots

- 
Screen Shot 2022-01-18 at 1 44 37 PM

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/redux/services/helm.ts
```diff
@@ -28,6 +29,15 @@ export function isHelmChartFolder(files: string[]) {
   return files.indexOf('Chart.yaml') !== -1 && files.indexOf('values.yaml') !== -1;
 }
 
+/**
+ * check if the k8sResource is supported
+ * @param resource
+ * @returns @boolean
+ */
+function isSupportedHelmResource(resource: K8sResource): boolean {
+  return !JSON.stringify(resource.content).includes('{"{');
```
olensmar, line 38: you can use resource.text instead - which won't require you to stringify. Also perhaps we should use a regex to check for a matching closing pair? i.e. '{{.*}}'

mortada-codes: resolved

src/redux/services/helm.ts
```diff
@@ -35,7 +35,9 @@ export function isHelmChartFolder(files: string[]) {
  * @returns @boolean
  */
 function isSupportedHelmResource(resource: K8sResource): boolean {
-  return !JSON.stringify(resource.content).includes('{"{');
+  const helmVariableRegex = /{"{/g;
```
olensmar, line 38: shouldn't this just be `{{.*}}` ?

mortada-codes: yeah, I am fixing it

mortada-codes: it is good now

devcatalin (approved):
LGTM

olensmar (approved):

mortada-codes:
should i update the test cases before merge?

olensmar:
> should i update the test cases before merge?

not sure - do they need updating?

olensmar:
then you should update them accordingly!

olensmar (approved):

PR #1043, opened 2022-01-19
network policy form editor & script for other resources

andreiv1992:
## Changes

- NetworkPolicy form schema and a script to generate form schemas for other resources

## Fixes

- NetworkPolicy form schema

## How to test it

- Open a network policy file and in the editor you should see a tab where you can edit network policy items in a form

## Screenshots

- https://user-images.githubusercontent.com/5509112/150130162-5543f723-1d4e-4236-b895-dd08073f3dfe.mov

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/scripts/generate-schema.ts
+++ b/src/scripts/generate-schema.ts
@@ -1,6 +1,8 @@
 import fs from 'fs';
+import path from 'path';
 
-const k8Path = `${__dirname}/../../resources/schemas/k8sschemas.json`;
+const k8Path = `${__dirname}${path.sep}..${path.sep}..${path.sep}resources${path.sep}schemas${path.sep}k8sschemas.json`;
+const schemaDir = `${__dirname}${path.sep}..${path.sep}..${path.sep}resources${path.sep}form-schemas`;
 
 const rootPropertiesToIgnore = ['apiVersion', 'kind', 'metadata'];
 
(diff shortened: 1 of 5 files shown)


andreiv1992:
If anyone thinks it's worth it I can add documentation for the script and make it more like a CLI, or maybe it should be moved to a separate repo.

src/scripts/generate-schema.ts
```diff
@@ -0,0 +1,122 @@
+import fs from 'fs';
+
+const k8Path = `${__dirname}/../../resources/schemas/k8sschemas.json`;
```
olensmar, line 3: use path.sep so this works on windows also!?

olensmar:
> If anyone thinks it's worth it I can add documentation for the script and make it more like a CLI, or maybe it should be moved to a separate repo.

olensmar:
sorry - pressed the wrong button...

src/scripts/generate-schema.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+  });
+
+  return {
+    properties: schemaDefinition,
+    propertiesUi: schemaUiDefinition,
+  };
+};
+
+generateSchema('NetworkPolicy');
```
olensmar, line 122: very cool! creating a generic cli sounds like a good idea - it should be able to handle any CRD - not just the bundled k8s ones!?

src/scripts/generate-schema.ts
```diff
@@ -0,0 +1,122 @@
...
+      }
+
+      const { properties, propertiesUi } = parseFields(kubernetesSchema[childKey].properties);
+      schemaDefinition[key] = {
+        type: 'object',
+        properties,
+      };
+      schemaUiDefinition[key] = {
+        'ui:title': `Object ${key}s`,
+        ...propertiesUi,
```
olensmar, line 78: perhaps you could add `'ui:widget' : 'namespaceSelection' ` to properties names "namespace" ?

src/scripts/generate-schema.ts
andreiv1992: Yes, as long as it has the same structure

src/scripts/generate-schema.ts
andreiv1992: Added

olensmar (approved):

olensmar:
Nice one @andreiv1992 !

PR #1044, opened 2022-01-19
feat: support for vanilla templates

devcatalin:
This PR...

## Changes

- adds support for vanilla templates

## Fixes

-

## How to test it

- Install the https://github.com/devcatalin/monokle-test-plugin which now contains a vanilla template as well

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/models/template.ts
+++ b/src/models/template.ts
@@ -9,7 +9,6 @@ const TemplateFormRuntype = Rt.Record({
 
 const TemplateManifestRuntype = Rt.Record({
   filePath: Rt.String,
-  fileRenameRule: Rt.String,
 });
 
 const TemplateBaseRuntype = Rt.Record({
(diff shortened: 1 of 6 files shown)


src/models/template.ts
```diff
@@ -9,7 +9,6 @@ const TemplateFormRuntype = Rt.Record({
 
 const TemplateManifestRuntype = Rt.Record({
   filePath: Rt.String,
-  fileRenameRule: Rt.String,
```
olensmar, line 12: wouldn't this be good to have here as well?

devcatalin: We can reintroduce it later if we allow saving the template output directly to files.
But for now, we're just creating unsaved resources from templates so this field is not needed

olensmar (approved):
lgtm!

PR #1053, opened 2022-01-19
fix: truncate title + fix width open button

topliceanurazvan:
## Changes

- Open button with fix width

## Fixes

- Truncate title if it's too long

## Screenshots

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
Looks good - but can we use a different label instead of "Open"? Perhaps "Use Template" ?  @tommywhitehead

devcatalin (approved):
Let's merge this and change the labels later, maybe after the next UX meeting

PR #1056, opened 2022-01-20
feat: Add Create Project menu to application menu

erdkse:
This PR...

## Changes

- Add Create project menu to application menu

## Fixes

-

## How to test it

-

## Screenshots

Screen Shot 2022-01-20 at 03 31 18

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan (commented):
If I have a project selected and I am not in the `Start project pane`, when I try to open a new project with `Select folder` it does nothing and if I go to the start project pane it automatically opens the folder explorer. 
Maybe we should show only a `New project` option that takes the user to the start project pane and from there he can select one of the options so it's consistent with the `New project` option from the project selector dropdown?

erdkse:
> If I have a project selected and I am not in the `Start project pane`, when I try to open a new project with `Select folder` it does nothing and if I go to the start project pane it automatically opens the folder explorer. Maybe we should show only a `New project` option that takes the user to the start project pane and from there he can select one of the options so it's consistent with the `New project` option from the project selector dropdown?

Thank you @topliceanurazvan for reporting - I think I have test it for this case but double-checking and fixing now

electron/menu.ts
```diff
@@ -92,6 +99,30 @@ const fileMenu = (state: RootState, dispatch: MainDispatch): MenuItemConstructor
         },
       },
       {type: 'separator'},
+      {
+        label: 'New Project',
+        submenu: [
+          {
+            label: 'Select Folder',
+            click: () => {
+              dispatch(openFolderExplorer());
```
olensmar, line 108: I think Monokle needs to run these actions in the renderer thread - you'll need to send an event via ipc to trigger this instead of dispatching them to the store directly

olensmar (changes requested):
see comment..

electron/menu.ts
erdkse: This dispatch is coming from `main.ts`. And It sends. the action to the current window. 

https://github.com/kubeshop/monokle/blob/main/electron/main.ts#L308

erdkse: but anyway it should send it to focused window - fixing it

olensmar: ok thanks - my bad! Shouldn't we be using this dispatch when selecting an existing project then?

olensmar:
> If I have a project selected and I am not in the `Start project pane`, when I try to open a new project with `Select folder` it does nothing and if I go to the start project pane it automatically opens the folder explorer. Maybe we should show only a `New project` option that takes the user to the start project pane and from there he can select one of the options so it's consistent with the `New project` option from the project selector dropdown?

I think that's an unnecessary step for the user - would prefer for these actions to work "directly".. @tommywhitehead  ?

olensmar:
> 

I get the same issue as reported by @topliceanurazvan

erdkse:
> > 
> 
> I get the same issue as reported by @topliceanurazvan

yeah me too - fixing

olensmar (approved):

electron/menu.ts
erdkse: This dispatch is for selecting an existing folder - no matter if it is an existing project or not - If the path is bound with an existing project then it will open it otherwise it will create it.

erdkse: If there is a problem with it I will prefer to fix it alongside fixing the non-existing path issue. Wdyt @olensmar ?

PR #1086, opened 2022-01-21
refactor: save duplicate resources with suffix

topliceanurazvan:
## Changes

- No longer replace duplicate resources, but save to files with suffixes.

## Screenshots

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
wondering if we could suggest different filenames based on some other algorithm - for example add the resource-kind - before resorting to numbers!? i.e. if Monokle detects multiple objects with the same name it automatically attempts to generate filenames  with  name-kind.yaml !?

topliceanurazvan:
@olensmar - great suggestion. We can definitely try to do that.

topliceanurazvan:
@olensmar - We are now generating filenames including the kind. PR is ready to be reviewed.

olensmar (approved):
nice!

PR #1098, opened 2022-01-22
Olelensmar/feat/form widgets

olensmar:
This PR adds form widgets for selecting resources, apiGroups and labels which are used by the default templates plugin. It also ensures that newly created resources are processed for links/refs.

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


erdkse (approved):
LGTM! Maybe we can create a base component and reuse some stuff since we are doing similar stuff in each components

olensmar:
> LGTM! Maybe we can create a base component and reuse some stuff since we are doing similar stuff in each components

thanks @erdkse - yes, we can probably consolidate these into a shared component with some properties/callback(s) for specific behavior

PR #1103, opened 2022-01-24
fix: small UI style bugs

erdkse:
This PR...

## Changes

- 

## Fixes

- Make project name ellipsis if overflow on recent projects and project selection

## How to test it

-

## Screenshots

Screen Shot 2022-01-24 at 09 42 01

Screen Shot 2022-01-24 at 09 42 40

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/PageHeader/ProjectSelection.tsx
+++ b/src/components/organisms/PageHeader/ProjectSelection.tsx
@@ -4,8 +4,6 @@ import {useSelector} from 'react-redux';
 import {Tooltip} from 'antd';
 import Column from 'antd/lib/table/Column';
 
-import {DownOutlined} from '@ant-design/icons';
-
 import _ from 'lodash';
 import {DateTime} from 'luxon';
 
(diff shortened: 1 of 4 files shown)


topliceanurazvan (approved):
Nice improvement by moving styled components to external file. Maybe we should remove the `ProjectButton` and have a `Container` instead so we remove the blue color on hover and focus (which makes it harder to read the name), but we can do that in another PR. LGTM!

erdkse:
> Nice improvement by moving styled components to external file. Maybe we should remove the `ProjectButton` and have a `Container` instead so we remove the blue color on hover and focus (which makes it harder to read the name), but we can do that in another PR. LGTM!

We have that improvement on Mahmoud - I didn't want to change things dramatically so he wouldn't receive conflicts - I am merging this PR so we can do it with another PR - Thank you for reviewing!!

PR #1133, opened 2022-01-25
fix: section quick create button color

devcatalin:
This PR...

## Changes

-

## Fixes

From:
image

To:
image

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar (approved):
LGTM - but to be honest I often mistakenly click on this button when I want to expand a section - perhaps it was a bad idea (of mine..) to put it here in the first place!?

devcatalin:
it also might be confusing to have two Plus buttons like in the screenshots above?
we could remove it entirely..

olensmar:
> it also might be confusing to have two Plus buttons like in the screenshots above? we could remove it entirely..

exactly @tommywhitehead what do you suggest? can we make it different so we don't have two plus-buttons? or should we remove?

olensmar:
I think we can merge this and discuss on the UX call if we should keep it!?

PR #1135, opened 2022-01-25
fix: hide left menu icons badges after pane opens

devcatalin:
This PR...

## Changes

- the blue dot/badge should be hidden after we click on the kustomization / helm charts icons
- it should appear again when changing the folder, if there are any kustomizations or helm charts in that folder

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/PaneManager/PaneManager.tsx
+++ b/src/components/organisms/PaneManager/PaneManager.tsx
@@ -1,4 +1,4 @@
-import React, {useContext, useMemo} from 'react';
+import React, {useContext, useEffect, useMemo, useState} from 'react';
 import {useSelector} from 'react-redux';
 
 import {Badge, Button, Skeleton, Space, Tooltip} from 'antd';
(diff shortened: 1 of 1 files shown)


erdkse (approved):
LGTM!

olensmar:
actually I don't agree with this - why would we hide it? I think it's helpful to have it there even after I've opened those panes (i'm forgetful..)

devcatalin:
> actually I don't agree with this - why would we hide it? I think it's helpful to have it there even after I've opened those panes (i'm forgetful..)

I was thinking it's just a hint that there are kustomizations in the pane, I'm expecting to see something "new" when I see that badge. If I remember correctly, this is how @tommywhitehead said we should do it..

olensmar:
ok let's merge and discuss with @tommywhitehead

PR #1202, opened 2022-01-27
fix: close dropdown on cluster select

topliceanurazvan:
## Fixes

- Closes the dropdown when a cluster is selected
- Handle cluster change only if the selected value is different than the current context

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar (approved):
great!

erdkse (approved):
lgtm!

Issue #1218, opened 2022-01-28
form editor modifies resources without it been opened


olensmar:
After using the form-editor the first time it seems to always be invoked in the background when navigating to other resources source editor - resulting in insertion of default values, etc (which is ok when the user actually uses the form editor).

PR #1225, opened 2022-01-28
fix: update active forms

mortada-codes:
This PR...

## Changes

- added an `isActive` flag to the FormEditor to only update active forms 

## Fixes

- form editor modifies resources without it been opened #1218 

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/FormEditor/FormEditor.tsx
+++ b/src/components/molecules/FormEditor/FormEditor.tsx
@@ -106,13 +106,8 @@ const FormContainer = styled.div`
   }
 `;
 
-interface IProps {
-  formSchema: any;
-  formUiSchema: any;
-  isActive: boolean;
-}
-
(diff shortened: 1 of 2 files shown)


mortada-codes:
I tried to use `<Fragment key="resourceId" />`  to let react manage the cycle by itself but i got error from Monaco editor
it would be great to refactor Action tabs in 1.6.0. because adding an isActive flag is an extra dependency that we can drop

olensmar:
thanks @mortada-codes - could we just render an empty panel if isActive is set to false!?

mortada-codes:
> thanks @mortada-codes - could we just render an empty panel if isActive is set to false!?

That's perfect

olensmar (approved):

PR #1230, opened 2022-01-30
fix: Kustomize schema/forms and errors related to conflicts with Flux

olensmar:
This PR fixes the form editor for Kustomizations and improves handling of both native and external Kustomization resources 

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/kindhandlers/common/customObjectKindHandler.ts
```diff
@@ -32,8 +33,8 @@ function extractResourceVersion(resource: K8sResource, kindVersion: string, kind
   return {version, group};
 }
 
-function extractFormSchema(editorSchema: any) {
-  const schema: any = JSON.parse(JSON.stringify(editorSchema));
+export function extractFormSchema(editorSchema: any) {
+  const schema: any = cloneDeep(editorSchema);
```
devcatalin, line 37: I would argue that cloning with JSON is faster based on some benchmarks I saw, but this is more readable

devcatalin (approved):
LGTM

src/kindhandlers/common/customObjectKindHandler.ts
olensmar: interesting - i thought it would be the opposite! thanks!

PR #1249, opened 2022-02-01
feat: Upgrade of CRA and Craco

DmitryEremenko:
## Changes

- Upgrade of CRA to version 5

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
I get the following after startup and every hot-redeploy - after closing the message Monokle seems to work ok.. 

image

erdkse:
> I get the following after startup and every hot-redeploy - after closing the message Monokle seems to work ok..
> 
> image
> 
> perhaps there is some configuration to show these errors in the console log only? (I think we had them before this upgrade also)

I solved this error for ever I believe 😄

PR #1259, opened 2022-02-01
fix: create namespace only if deploy is successful

topliceanurazvan:
## Fixes

- Delete the newly created namespace if the deploy fails

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar (approved):
LGTM!

Issue #731, opened 2021-11-20
show context-sensitive documentation next to editor


olensmar:
to help me understand the properties in my k8s resources I can currently hover over them to see a popup with context-sensitive documentation - I would like to see this while i'm moving around with the caret instead of having to hover and wait for the popup - perhaps in a tab below the editor!? as an additional tab to what is suggested in #367

devcatalin:
It would be nice to show what properties can be added while moving the text cursor around through objects in a manifest file

tommywhitehead:
<img width="1440" alt="#731 show context-sensitive documentation next to editor 1" src="https://user-images.githubusercontent.com/4236167/147701672-b9b820ee-1c5f-4456-bba2-7a1611a32e49.png">

Figma:
https://www.figma.com/file/PGd55pajyuL0u8ydVQtOuh/%23731-show-context-sensitive-documentation-next-to-editor?node-id=0%3A1

PR #1278, opened 2022-02-02
refactor: actions pane footer

topliceanurazvan:
## Changes

- Redesign of the actions pane footer
- Make actions pane footer resizable 

## Screenshots

image

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/atoms/SplitView/SplitView.tsx
+++ b/src/components/atoms/SplitView/SplitView.tsx
@@ -491,7 +491,7 @@ const SplitView: FunctionComponent<SplitViewProps> = ({
         <StyledDivider />
       </StyledDividerHitBox>
 
-      <Pane id="EditorPane" width={editWidth * viewWidth} hide={false}>
+      <Pane id="EditorPane" width={editWidth * viewWidth + 5} hide={false}>
         {editor}
       </Pane>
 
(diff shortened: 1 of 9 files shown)


olensmar:
Looks good - although I thought the Terminal was going to span the entire window!? (maybe this is just an example)

topliceanurazvan:
@olensmar - the terminal it's just an example, this was made after the design from https://github.com/kubeshop/monokle/issues/731 , we can show there any content like `Documentation`, `Terminal` etc.. 

Maybe we can have multiple tabs in this footer that could be useful for the user. @devcatalin @tommywhitehead wdyt?

olensmar:
> @olensmar - the terminal it's just an example, this was made after the design from #731 , we can show there any content like `Documentation`, `Terminal` etc..
> 
> Maybe we can have multiple tabs in this footer that could be useful for the user. @devcatalin @tommywhitehead wdyt?

thanks - that's what I thought! Definitely agree on multiple tabs!

erdkse:
I prefer the multiple tabs to be on an entire pane - not separated panes for each

devcatalin (approved):
This works nicely, good job!

PR #1283, opened 2022-02-02
feat: add Preview action for file explorer

DmitryEremenko:
## Changes

- Add Preview action for file explorer

## How to test it

- Open any project, find something that can be previewed , hover over and run Preview

## Screenshots
image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/FileTreePane/TreeItem.tsx
```diff
@@ -255,8 +255,8 @@ export const TreeItem: React.FC<TreeItemProps> = props => {
       ) : null}
       {isTitleHovered && !processingEntity.processingType ? (
         <>
-          {canPreview(relativePath) && !isInPreviewMode && (
```
topliceanurazvan: Should we see the Preview label when in preview mode? The file explorer is disabled and it's confusing if you can still interact with the `Preview` action. Maybe make it disabled or just don't show it?

DmitryEremenko: In Preview mode we also can access sub menu of item, should I hide it too then?

olensmar: i think the general approach is to disable - not to hide - @tommywhitehead ?

DmitryEremenko: I've disabled this button in Preview mode, "..." still allow to click on it in preview mode. 
Intention of this feature is to mark file as something that we can preview in Navigator

src/components/organisms/FileTreePane/TreeItem.tsx
```diff
@@ -0,0 +1,282 @@
...
+        </S.SpinnerWrapper>
+      ) : null}
+      {isTitleHovered && !processingEntity.processingType ? (
+        <>
+          {canPreview(relativePath) && (
+            <S.PreviewSpan
+              type="text"
+              size="small"
+              disabled={isInPreviewMode}
+              isItemSelected={isFileSelected}
```
topliceanurazvan, line 263: We should put `$` in front of isItemSelected prop used for styling in order to avoid React warnings.

olensmar:
LGTM but I think the original purpose of the issue got lost - I still have to hover each item to know if it is previewable - i was hoping for some kind of indication without that requirement - @tommywhitehead ?

devcatalin:
Maybe an antd Tag next to the filename that says "Previewable" ?

tommywhitehead:
Maybe we can use a 12px `gray-7` icon = `<EyeOutlined />` as part of the label (directly to the right of the name of the resource) to indicate that it is previewable.

Like this:
image

olensmar:
> Maybe we can use a 14px `gray-7` icon = `<EyeOutlined />` as part of the label (directly to the right of the name of the resource) to indicate that it is previewable.
> 
> Like this: image

thanks @tommywhitehead - let's try that

src/components/organisms/FileTreePane/TreeItem.tsx
DmitryEremenko: Done!

olensmar (approved):
great - thank you!

PR #1291, opened 2022-02-03
fix Helm chart folder recognition

devcatalin:
This PR...

## Changes

- added option to show empty section groups if there's an `emptyGroupText` property set
- correctly show charts for which we did not find any values files

## Fixes

- isHelmChartFolder to only check for `Chart.yaml` file

## How to test it

-

## Screenshots

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/navsections/sectionBlueprintMiddleware.ts
+++ b/src/navsections/sectionBlueprintMiddleware.ts
@@ -274,7 +274,9 @@ const processSectionBlueprints = async (state: RootState, dispatch: AppDispatch)
       visibleItemIds: g.itemIds.filter(itemId => itemInstanceMap[itemId].isVisible === true),
     }));
     const visibleItemIds = itemInstances?.filter(i => i.isVisible === true).map(i => i.id) || [];
-    const visibleGroupIds = sectionInstanceGroups.filter(g => g.visibleItemIds.length > 0).map(g => g.id);
+    const visibleGroupIds = sectionBlueprint.customization?.emptyGroupText
+      ? sectionInstanceGroups.map(g => g.id)
+      : sectionInstanceGroups.filter(g => g.visibleItemIds.length > 0).map(g => g.id);
     const sectionInstance: SectionInstance = {
       id: sectionBlueprint.id,
       rootSectionId: sectionBlueprint.rootSectionId,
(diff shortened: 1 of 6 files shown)


src/navsections/sectionBlueprintMiddleware.ts
```diff
@@ -274,7 +274,9 @@ const processSectionBlueprints = async (state: RootState, dispatch: AppDispatch)
       visibleItemIds: g.itemIds.filter(itemId => itemInstanceMap[itemId].isVisible === true),
     }));
     const visibleItemIds = itemInstances?.filter(i => i.isVisible === true).map(i => i.id) || [];
-    const visibleGroupIds = sectionInstanceGroups.filter(g => g.visibleItemIds.length > 0).map(g => g.id);
+    const visibleGroupIds = sectionBlueprint.customization?.emptyGroupText
+      ? sectionInstanceGroups.map(g => g.id)
+      : sectionInstanceGroups.filter(g => g.visibleItemIds.length > 0).map(g => g.id);
```
DmitryEremenko, line 279: You can reduce this line by using reduce() method for this array, in that whay you will run only once through array

src/navsections/sectionBlueprintMiddleware.ts
```diff
@@ -274,7 +274,9 @@ const processSectionBlueprints = async (state: RootState, dispatch: AppDispatch)
       visibleItemIds: g.itemIds.filter(itemId => itemInstanceMap[itemId].isVisible === true),
     }));
     const visibleItemIds = itemInstances?.filter(i => i.isVisible === true).map(i => i.id) || [];
```
DmitryEremenko, line 276: Here too as in comment below

olensmar (approved):

DmitryEremenko (changes requested):

src/navsections/sectionBlueprintMiddleware.ts
devcatalin: I decided to go with a lodash sequence instead to keep the readability and also loop through the array only once

DmitryEremenko (approved):
LGTM

PR #1301, opened 2022-02-04
Razvantopliceanu/feat/xterm terminal

topliceanurazvan:
## Changes

- Add xterm.js basic terminal, still needs lots of improvements/polishing

TODO:
   - Find a way to kill the pty process when the terminal tab is closed or use the same instance?
   - Make the terminal resizable relative to the parent container ( maybe use it's cols and rows options or resize() )

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


andreiv1992:
It would be useful if we can add some basic tests around this, opening the terminal, writing a simple command and closing it

olensmar:
agree with @andreiv1992 - although I think it can wait until we decide to actually implement this

WitoDelnat:
I wonder whether we should close this draft for the time being until its picked up again?

topliceanurazvan:
@WitoDelnat I think we can close it as it is deprecated and has a lot of conflicts. Could be used for reference when the issue will be picked up again. @devcatalin wdyt?

PR #1304, opened 2022-02-05
feat: add confirm dialog on window close with unsaved files

### LINKED ISSUE #991: Show confirm dialog for unsaved resources
It would be nice to show a confirm dialog when the application is closed and still has unsaved resources.

It should probably say something like this: "You have X resources unsaved. By closing Monokle you will lose your progress."

WitoDelnat:
This PR resolves #991

## Changes

- It now shows a confirmation dialog when you close the window with unsaved files.

## Fixes

- none

## How to test it

- Create a new resource and be sure to use the default "Don't save".
- Close the window or quit the application.

## Screenshots

Screenshot 2022-02-05 at 10 57 58

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -13,7 +13,7 @@ moduleAlias.addAliases({
   '@root': `${__dirname}/../`,
 });
 
-import {app, BrowserWindow, nativeImage, ipcMain, dialog} from 'electron';
+import {app, BrowserWindow, nativeImage, ipcMain} from 'electron';
 import * as path from 'path';
 import installExtension, {REACT_DEVELOPER_TOOLS, REDUX_DEVTOOLS} from 'electron-devtools-installer';
 import * as Splashscreen from '@trodi/electron-splashscreen';
@@ -64,7 +64,7 @@ import {AnyTemplate, TemplatePack} from '@models/template';
 import {AnyPlugin} from '@models/plugin';
 import {AnyExtension, DownloadPluginResult, DownloadTemplatePackResult, DownloadTemplateResult, UpdateExtensionsResult} from '@models/extension';
 import {KustomizeCommandOptions} from '@redux/thunks/previewKustomization';
-import { convertRecentFilesToRecentProjects, getSerializedProcessEnv, saveInitialK8sSchema, setProjectsRootFolder } from './utils';
+import { askActionConfirmation, convertRecentFilesToRecentProjects, getSerializedProcessEnv, saveInitialK8sSchema, setProjectsRootFolder } from './utils';
 import {InterpolateTemplateOptions} from '@redux/services/templates';
 import {StartupFlags} from '@utils/startupFlag';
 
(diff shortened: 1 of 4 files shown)


WitoDelnat:
I wanted to explore the Monokle codebase and I prefer a hands-on approach. I only tested this on MacOS and since this native behaviour can change between OS it might be good for someone with Windows to test it as well.

Let me know if I need to change anything to comply with contributing standards.

electron/main.ts
```diff
@@ -368,6 +370,30 @@ export const createWindow = (givenPath?: string) => {
 
   });
 
+  win.on('close', (e) => {
+    if (unsavedResourceCount === 0) {
+      return;
+    }
+
+    const message = unsavedResourceCount === 1
+      ? 'You have an unsaved resource'
```
WitoDelnat, line 379: I wonder if unsaved files count makes more sense than unsaved resources count? You can have multiple unsaved resources in one file which doesn't give the user clear info on how many actions he has to perform before the warning disappears.

Example
- 2 unsaved files -> you close one and you know you still need to close another.
- 5 unsaved resources across 2 files -> you close one and don't know how many files you still need to find and save afterwards.

devcatalin: Thanks for your contribution and good job, this looks great!
To answer you question, we only have the unsaved state for resources that are not already saved in a file. File changes are automatically saved.
Thus, using the unsaved resources count like you already did is the way to go.

devcatalin:
We will also have to show a similar modal when the user changes the current active project (using the project selector from the header), since the changes will be lost in that case as well.
Let us know if you'd like to do that as well, or we can make a separate issue/PR.

electron/main.ts
WitoDelnat: That makes sense, thanks for the clarification!

WitoDelnat:
Yeah, no problem. I like to finish what I've started and I already have a solution in mind. I'm a bit occupied this weekend so I'll give it a go later this week.

chlete:
Hi @WitoDelnat ! Thanks a lot! I'm Christian, the product manager of Monokle. We really appreciate your contribution ⭐ !

WitoDelnat:
Hi @chlete! Nice to meet you.

@devcatalin I've updated the MR with a similar modal when the user changes the current active project. Once again, I only tested this on MacOS.

src/components/organisms/PageHeader/ProjectSelection.tsx
```diff
@@ -78,6 +81,15 @@ const ProjectSelection = () => {
   };
 
   const handleProjectChange = (project: Project) => {
+    const confirmed = ipcRenderer.sendSync('confirm-action', {
```
WitoDelnat, line 84: I imagine this is more than fast enough since the IPC happens on the same computer but my experience with Electron is limited. We can optimise a bit by doing the skip if unsaved resource count equals zero here but I left it lean until performance is an issue.

WitoDelnat:
Hi @devcatalin! I saw that this MR is still open. Please let me know if anything is needed from my side. I already quickly rebased and resolved the merge conflicts.

devcatalin (approved):
Looks good to me!

PR #1306, opened 2022-02-07
fix: ensured passing of main process env to kubeclient exec authenticator

### LINKED ISSUE #1285: Use Paths for aws cli or aws-iam-authenticator for EKS AuthN - MacOS
**Describe the enhancement you'd like to see**
In my `.kube/config` I use either `aws eks get-token` or `aws-iam-authenticator` as the commands to authenticate to EKS clusters.

```
...
      args:
      - token
      - -i
      - monokle
      command: aws-iam-authenticator
...
```

or

```
...
      command: aws
      args:
        - "eks"
        - "get-token"
        - "--cluster-name"
        - "monokle"
 ...
```

The tools that create the `user` elements in the `.kube/config` do not include the absolute paths for these binaries. My shell resolves the paths. It would be very handy for Monokle to resolve the paths as well. This would reduce troubleshooting, and the need to continually edit `kube/config` files `user` elements.

Without the absolute paths (`which aws` or `which aws-iam-authenticator`) the AuthN fails with the following error:
```
Error: Command failed with ENOENT: aws --region <REGION> eks get-token --cluster-name <CLUSTER_NAME> spawnSync aws ENOENT
```

olensmar:
This PR ensures that all kubeclient instances are created consistently - and ensures that kubeconfig users with exec authentication use the main process env for exec call

## Changes

-

## Fixes

#1285 

## How to test it

- create a context in kubeconfig that has a user that uses exec authentication (aws) - and attempt to preview this context in Monokle

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/redux/reducers/main.ts
+++ b/src/redux/reducers/main.ts
@@ -195,12 +195,6 @@ function getLocalResourceMap(state: Draft<AppState>) {
   return localResourceMap;
 }
 
-let mainProcessEnv: any | undefined;
-
-export function getMainProcessEnv() {
-  return mainProcessEnv;
-}
-
 /**
  * The main reducer slice
  */
@@ -209,9 +203,6 @@ export const mainSlice = createSlice({
   name: 'main',
   initialState: initialState.main,
   reducers: {
-    setMainProcessEnv: (state: Draft<AppState>, action: PayloadAction<any>) => {
-      mainProcessEnv = action.payload;
-    },
     setAppRehydrating: (state: Draft<AppState>, action: PayloadAction<boolean>) => {
       state.isRehydrating = action.payload;
       if (!action.payload) {
@@ -1130,6 +1121,5 @@ export const {
   addMultipleKindHandlers,
   addKindHandler,
   seenNotifications,
-  setMainProcessEnv,
 } = mainSlice.actions;
 export default mainSlice.reducer;
--- a/src/redux/reducers/main.ts
+++ b/src/redux/reducers/main.ts
@@ -38,6 +38,7 @@ import {setRootFolder} from '@redux/thunks/setRootFolder';
 
 import electronStore from '@utils/electronStore';
 import {getFileStats, getFileTimestamp} from '@utils/files';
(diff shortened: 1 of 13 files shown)


src/redux/reducers/main.ts
```diff
@@ -202,6 +209,9 @@ export const mainSlice = createSlice({
   name: 'main',
   initialState: initialState.main,
   reducers: {
+    setMainProcessEnv: (state: Draft<AppState>, action: PayloadAction<any>) => {
+      mainProcessEnv = action.payload;
```
devcatalin, line 213: this is a bit weird, should we live it like this for now?

the alternative would be to create an ipc event and send that from main process instead of dispatching this action

src/utils/kubeclient.ts
```diff
@@ -0,0 +1,55 @@
...
+  const path = typeof config === 'string' ? config : config.projectConfig?.kubeConfig?.path || config.kubeConfig.path;
+  if (!path) {
+    throw new Error('Missing path to kubeconfing');
+  }
+
+  kc.loadFromFile(path);
+  let currentContext =
+    context || typeof config === 'string'
+      ? undefined
+      : config.projectConfig?.kubeConfig?.currentContext || config.kubeConfig.currentContext;
```
devcatalin, line 21: shouldn't this be:
```  
let currentContext = typeof config === 'string'
      ? context
      : config.projectConfig?.kubeConfig?.currentContext || config.kubeConfig.currentContext;
```

src/redux/reducers/main.ts
olensmar: that's what I had initially - i preferred this approach - but maybe an IPC event is easier to understand!?

src/utils/kubeclient.ts
olensmar: yes - thanks

devcatalin (approved):
LGTM

PR #1318, opened 2022-02-08
Support multiple k8s versions

erdkse:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/SettingsManager/Settings.tsx
```diff
@@ -228,6 +264,34 @@ export const Settings = ({
           </>
         </Form>
       )}
+      <S.Div>
+        <S.Span>K8S Version</S.Span>
+        <div>
+          <Tooltip title="">
```
olensmar, line 270: missing tooltip - suggest "Select which kubernetes schema version to use for validation"

olensmar:
looks great @erdkse - added some minor comments..

since userDataDir never changes i wonder if we should just set it as a global constant instead of passing it around? (same goes for the other dirs..) - this could be a separate refactoring that we could do "later"

src/redux/services/schema.ts
```diff
@@ -28,10 +26,23 @@ const HELM_CHART_SCHEMA = JSON.parse(loadResource('form-schemas/helm-chart-schem
...
+    k8sSchema = k8sSchemaCache.get(schemaVersion);
+  } else {
+    // @ts-ignore
+    k8sSchema = JSON.parse(
+      readFileSync(path.join(String(userDataDir), path.sep, 'schemas', `${schemaVersion}.json`), 'utf-8')
+    );
+    k8sSchemaCache.set(schemaVersion, k8sSchema);
+  }
+
   if (isKustomizationResource(resource)) {
```
olensmar, line 46: move this up so we don't get the k8sSchema unnecessarily

src/redux/services/schema.ts
```diff
@@ -28,10 +26,23 @@ const HELM_CHART_SCHEMA = JSON.parse(loadResource('form-schemas/helm-chart-schem
 // @ts-ignore
 const HELM_CHART_UI_SCHEMA = JSON.parse(loadResource('form-schemas/helm-chart-ui-schema.json'));
 
+let k8sSchemaCache = new Map<string, any | undefined>();
```
olensmar, line 29: const ?

olensmar (approved):
LGTM!

src/redux/services/schema.ts
```diff
@@ -44,6 +44,7 @@ export function getResourceSchema(resource: K8sResource, schemaVersion: string,
     k8sSchema = JSON.parse(
       readFileSync(path.join(String(userDataDir), path.sep, 'schemas', `${schemaVersion}.json`), 'utf-8')
     );
+    k8sSchemaCache = new Map<string, any | undefined>();
```
olensmar, line 47: k8SchemaCache.clear() is less drastic!? and you would have to clear the schemaCache also

PR #1325, opened 2022-02-09
fix: added more properties to ignore during validation

olensmar:
This PR adds the following properties to the list of ignored properties during validation:

- finishedAt
- createdAt
- startedAt
- x_kubernetes_preserve_unknown_fields

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/constants/clusterResource.ts
+++ b/src/constants/clusterResource.ts
@@ -1,5 +1,10 @@
 export const CLUSTER_RESOURCE_IGNORED_PATHS = [
   '...creationTimestamp',
+  '...lastProbeTime',
+  '...finishedAt',
+  '...createdAt',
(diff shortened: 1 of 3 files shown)


devcatalin:
@olensmar should we add all of these to the `CLUSTER_RESOURCE_IGNORED_PATHS` constant and vice versa?

olensmar:
> @olensmar should we add all of these to the `CLUSTER_RESOURCE_IGNORED_PATHS` constant and vice versa?

good catch - I wasn't aware of that - i'll add them there and use that list in the validation for now

Issue #586, opened 2021-11-04
Allow user to provide cmd-line arguments for Helm preview


olensmar:
When running a Helm preview I need to specify cmd-line options - which I currently can't - so the preview isn't working correctly. 

It would be nice to have a form where I can specify option values as described at https://helm.sh/docs/helm/helm_template/#options and https://helm.sh/docs/helm/helm_install/#options. The form would show check-boxes for boolean options, input fields for text, etc. 

My specified values would be remembered independently for each values file so when I preview again I don't have to specify the values again. A nive-to-have feature would be to have "preview profiles" allowing me to easily switch between different settings of options.

tommywhitehead:
Let's review these in tomorrow UX call:
https://www.figma.com/file/amLogk20OPgdpVbl7qlK3g/%23586-%23946-Multiple-values.yaml-file-preview-and-configuration?node-id=0%3A1

PR #1338, opened 2022-02-10
refactor: create subsections for helm charts

devcatalin:
In this PR I've created a "Values Files" subsection for each helm chart as a starting point for this issue - https://github.com/kubeshop/monokle/issues/586

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
looks good - but the numbering after the headlines is a bit confusing as it shows the total number of files in the hierarchy - perhaps we could change this so:
- the number after "Helm Charts" shows the number of Helm charts (which would be 6 in the screenshot)
- we don't show a number after name of the helm chart itself

olensmar (approved):
nice!

PR #1343, opened 2022-02-11
re-parsing of invalid resources improvements

olensmar:
This PR improves re-parsing of updated resources when they are missing kind/apiVersion

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/redux/services/resource.ts
```diff
@@ -457,8 +458,17 @@ export function reprocessResources(
         resource.name = `Patch: ${resource.name}`;
       }
 
-      resource.kind = resource.content.kind;
-      resource.version = resource.content.apiVersion;
+      const isKustomziationFile = resource.filePath.endsWith('kustomization.yaml');
```
devcatalin, line 461: let's have a KUSTOMIZATION_FILE_NAME constant here and use it in the `isKustomizationFile` from `kustomize.ts` as well

src/components/molecules/Monaco/Monaco.tsx
```diff
@@ -191,7 +188,8 @@ const Monaco = (props: {diffSelectedResource: () => void; applySelection: () =>
     if (selectedResourceId) {
       // this will slow things down if document gets large - need to find a better solution...
       const documents = parseAllDocuments(newValue);
-      setValid(documents.length > 0 && !documents.some(d => !isValidResourceDocument(d)));
+      // only accept single document changes for now
```
devcatalin, line 191: does editing multi-resource files still work in this case? if not, should the editor be read-only?

devcatalin (changes requested):

src/redux/services/resource.ts
olensmar: good idea!

src/components/molecules/Monaco/Monaco.tsx
olensmar: i'll double-check, but yes - selectedResourceId should not be set if the user is editing a multiresource file. This specific case is only to ensure that a user doesn't try to turn a single resources into multiple resources by manually inserting a resource delimiter (which we could support in the future)

devcatalin (approved):
Nice!

PR #1352, opened 2022-02-15
Olelensmar/feat/helm chart form editors

olensmar:
This PR adds form editors and source-editor validation for Helm Chart.yaml files and values files that have a corresponding values.schema.json file

Helm Chart schema and UI-schema descriptions were taken from https://helm.sh/docs/topics/charts/#the-chartyaml-file

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

image

image

image

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/resources/form-schemas/helm-chart-schema.json
+++ b/resources/form-schemas/helm-chart-schema.json
@@ -77,7 +77,6 @@
             {
               "type": "string"
             }
-
           },
           "import-values": {
             "type": "array",
--- a/src/components/molecules/FormEditor/FormEditor.tsx
+++ b/src/components/molecules/FormEditor/FormEditor.tsx
@@ -6,14 +6,17 @@ import {useDebounce} from 'react-use';
 import {Theme as AntDTheme} from '@rjsf/antd';
 import {withTheme} from '@rjsf/core';
 
+import fs from 'fs';
+import log from 'loglevel';
 import styled from 'styled-components';
-import {stringify} from 'yaml';
+import {parseDocument, stringify} from 'yaml';
 
 import {DEFAULT_EDITOR_DEBOUNCE} from '@constants/constants';
 
-import {useAppDispatch} from '@redux/hooks';
+import {useAppDispatch, useAppSelector} from '@redux/hooks';
 import {updateResource} from '@redux/reducers/main';
 import {isInPreviewModeSelector, selectedResourceSelector, settingsSelector} from '@redux/selectors';
+import {getAbsoluteFilePath} from '@redux/services/fileEntry';
 import {mergeManifests} from '@redux/services/manifest-utils';
 import {removeSchemaDefaults} from '@redux/services/schema';
 
@@ -110,6 +113,8 @@ const FormContainer = styled.div`
 const FormEditor = (props: {formSchema: any; formUiSchema?: any}) => {
   const {formSchema, formUiSchema} = props;
   const selectedResource = useSelector(selectedResourceSelector);
+  const selectedPath = useAppSelector(state => state.main.selectedPath);
+  const fileMap = useAppSelector(state => state.main.fileMap);
   const [formData, setFormData] = useState<any>();
   const dispatch = useAppDispatch();
   const isInPreviewMode = useSelector(isInPreviewModeSelector);
@@ -122,12 +127,23 @@ const FormEditor = (props: {formSchema: any; formUiSchema?: any}) => {
 
   useDebounce(
     () => {
-      if (selectedResource) {
-        let formString = stringify(formData);
-        const content = mergeManifests(selectedResource.text, formString);
+      let formString = stringify(formData);
 
+      if (selectedResource) {
+        const content = mergeManifests(selectedResource.text, formString);
         if (content.trim() !== selectedResource.text.trim()) {
           dispatch(updateResource({resourceId: selectedResource.id, content}));
+        }
+      } else if (selectedPath) {
+        try {
+          const filePath = getAbsoluteFilePath(selectedPath, fileMap);
+          const fileContent = fs.readFileSync(filePath, 'utf8');
+          const content = mergeManifests(fileContent, formString);
+          if (content.trim() !== fileContent.trim()) {
+            fs.writeFileSync(filePath, content);
+          }
+        } catch (e) {
+          log.error(`Failed to update file [${selectedPath}]`, e);
         }
       }
     },
@@ -138,9 +154,17 @@ const FormEditor = (props: {formSchema: any; formUiSchema?: any}) => {
   useEffect(() => {
     if (selectedResource) {
       setFormData(selectedResource.content);
+    } else if (selectedPath) {
+      try {
+        const filePath = getAbsoluteFilePath(selectedPath, fileMap);
+        const fileContent = fs.readFileSync(filePath, 'utf8');
+        setFormData(parseDocument(fileContent).toJS());
+      } catch (e) {
+        log.error(`Failed to read file [${selectedPath}]`, e);
+      }
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [selectedResource]);
+  }, [selectedResource, selectedPath]);
 
   useEffect(() => {
     if (!settings.createDefaultObjects || !settings.setDefaultPrimitiveValues) {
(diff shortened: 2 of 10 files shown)


src/redux/services/fileEntry.ts
```diff
@@ -195,6 +195,14 @@ export function getAbsoluteResourcePath(resource: K8sResource, fileMap: FileMapT
   return path.join(fileMap[ROOT_FILE_ENTRY].filePath, resource.filePath);
 }
 
+/**
+ * Returns the absolute path to the file that containing specified resource
```
devcatalin, line 199: this comment should be modified

src/redux/services/schema.ts
```diff
@@ -58,6 +68,24 @@ export function getResourceSchema(resource: K8sResource) {
   return schemaCache.get(resource.kind);
 }
 
+export function getSchemaForPath(filePath: string, fileMap: FileMapType): any | undefined {
+  if (filePath?.toLowerCase().endsWith('/chart.yaml')) {
```
devcatalin, line 72: should use `path.sep` here

src/hooks/useResourceYamlSchema.ts
```diff
@@ -30,7 +34,7 @@ function useResourceYamlSchema(yaml: any, resource: K8sResource | undefined) {
         enableSchemaRequest: true,
         hover: true,
         completion: true,
-        isKubernetes: true,
+        isKubernetes: resource,
```
devcatalin, line 37: should this be `Boolean(resource)` ? but it's maybe redundant

devcatalin: I just saw that `| undefined` was removed from the `resource` argument, so wouldn't this be always true?

src/components/organisms/ActionsPane/ActionsPane.tsx
```diff
@@ -341,7 +342,12 @@ const ActionsPane = (props: {contentHeight: string}) => {
   }, [selectedResourceId, resourceMap]);
 
   useEffect(() => {
```
devcatalin, line 344: missing `selectedPath` as dependency in this `useEffect`

src/redux/services/schema.ts
devcatalin: or use `path.basename(filePath).toLowerCase() === 'chart.yaml'`

olensmar:
now with a ui-schema for the helm-chart: 
image

src/components/organisms/ActionsPane/ActionsPane.tsx
olensmar: adding this breaks the source editor (it doesn't show up - not sure why)

resources/form-schemas/helm-chart-schema.json
```diff
@@ -0,0 +1,149 @@
...
+    },
+    "maintainers": {
+      "type": "array",
+      "items":
+      {
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string",
+            "description": "The maintainers name"
```
devcatalin, line 108: small type here and below, should be `maintainer's`

resources/form-schemas/helm-chart-schema.json
```diff
@@ -0,0 +1,149 @@
...
+          "import-values": {
+            "type": "array",
+            "description": "ImportValues holds the mapping of source values to parent key to be imported. Each item can be a string or pair of child/parent sublist items.",
+            "items":
+            {
+              "type": "string"
+            }
+          },
+          "alias": {
+            "description": "Alias to be used for the chart. Useful when you have to add the same chart multiple times",
```
devcatalin, line 90: what does "when you have to add the same chart multiple times" mean?

src/components/organisms/ActionsPane/ActionsPane.tsx
olensmar: sorry - my bad - you're right!

resources/form-schemas/helm-chart-schema.json
```diff
@@ -0,0 +1,149 @@
...
+            "type": "array",
+            "description": "Tags can be used to group charts for enabling/disabling together",
+            "items":
+            {
+              "type": "string"
+            }
+          },
+          "import-values": {
+            "type": "array",
+            "description": "ImportValues holds the mapping of source values to parent key to be imported. Each item can be a string or pair of child/parent sublist items.",
```
devcatalin, line 83: maybe this is easier to read: `ImportValues defines the source value to parent key mappings that will be imported.`

resources/form-schemas/helm-chart-schema.json
```diff
@@ -0,0 +1,149 @@
...
+            "type": "string",
+            "description": "The version of the chart (\"1.2.3\")"
+          },
+          "repository": {
+            "type": "string",
+            "description": "The repository URL (\"https://example.com/charts\") or alias (\"@repo-name\")"
+          },
+          "condition": {
+            "type": "string",
+            "description": "A yaml path that resolves to a boolean, used for enabling/disabling charts (e.g. subchart1.enabled )"
```
devcatalin, line 71: I suggest: `yaml object path`

resources/form-schemas/helm-chart-schema.json
```diff
@@ -0,0 +1,149 @@
...
+      {
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string",
+            "description": "The name of the chart (nginx)"
+          },
+          "version": {
+            "type": "string",
+            "description": "The version of the chart (\"1.2.3\")"
```
devcatalin, line 63: let's add `e.g.` at the start of the examples for all properties

resources/form-schemas/helm-chart-schema.json
```diff
@@ -0,0 +1,149 @@
...
+      }
+
+    },
+    "home": {
+      "type": "string",
+      "description": "The URL of this projects home page"
+    },
+    "sources": {
+      "type": "array",
+      "description": "A list of URLs to source code for this project",
```
devcatalin, line 44: I suggest: `list of URLs pointing to the source code of this...`

resources/form-schemas/helm-chart-schema.json
```diff
@@ -0,0 +1,149 @@
...
+      "description": "A list of keywords about this project",
+      "items":
+      {
+        "type": "string"
+      }
+
+    },
+    "home": {
+      "type": "string",
+      "description": "The URL of this projects home page"
```
devcatalin, line 40: small typo: `project's`

resources/form-schemas/helm-chart-schema.json
```diff
@@ -0,0 +1,149 @@
...
+      "type": "string",
+      "description": "A URL to an SVG or PNG image to be used as an icon"
+    },
+    "appVersion": {
+      "type": "string",
+      "description": "The version of the app that this contains. Needn't be SemVer. Quotes recommended."
+    },
+    "deprecated": {
+      "type": "boolean",
+      "description": "Whether this chart is deprecated"
```
devcatalin, line 134: add "or not" at the end?

src/components/molecules/FormEditor/FormEditor.tsx
```diff
@@ -122,25 +127,44 @@ const FormEditor = (props: {formSchema: any; formUiSchema?: any}) => {
...
       if (selectedResource) {
-        let formString = stringify(formData);
         const content = mergeManifests(selectedResource.text, formString);
-
         if (content.trim() !== selectedResource.text.trim()) {
           dispatch(updateResource({resourceId: selectedResource.id, content}));
         }
+      } else if (selectedPath) {
+        try {
+          const filePath = getAbsoluteFilePath(selectedPath, fileMap);
```
devcatalin, line 139: I think for future consideration we should have selectors for these types of functions so we don't have to import the entire `fileMap` state into each components that needs to call a simple function like this one.
By using custom selectors, only the output of the selector will affect the rendering of the component, while in this case, any fileMap change will trigger a re-render of this component

devcatalin (changes requested):

devcatalin (approved):
Looks really nice!

Issue #1374, opened 2022-02-16
When deploying a namespace, namespace selector is shown


aabedraba:
**Describe the bug**
When deploying a namespace, the prompt for choosing a namespace appears, although it does not affect the deployment. 

**To Reproduce**
1. Create a namespace through Monokle
2. Try to deploy it 
3. See a selector of "namespaces to deploy to" 
4. Any selection and further deployment does not affect deployment behavior

**Expected behavior**
No namespace selector shown. 

**Screenshots**
![image](https://user-images.githubusercontent.com/27779735/154354022-b9f5946f-68c9-4729-9d1d-f0516f8e41f3.png)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Version 1.5.1

**Additional context**

PR #1375, opened 2022-02-16
fix: disable namespace selection elements if a namespace is being dep…

olensmar:
…loyed

This PR...

## Changes

-

## Fixes

- #1374 

## How to test it

-

## Screenshots

image

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
+++ b/src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
@@ -97,7 +97,7 @@ const ModalConfirmWithNamespaceSelect: React.FC<IProps> = props => {
     } else if (selectedOption === 'none') {
       onOk();
     }
-  }, [kubeConfigContext, createNamespaceName, kubeConfigPath, selectedNamespace, selectedOption, onOk]);
+  }, [kubeConfigContext, createNamespaceName, kubeConfigPath, selectedNamespace, selectedOption, onOk, configState]);
 
   useEffect(() => {
     if (defaultOption && defaultOption === 'none') {
--- a/src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
+++ b/src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
@@ -1,6 +1,6 @@
 import * as k8s from '@kubernetes/client-node';
 
-import React, {useCallback, useEffect, useState} from 'react';
+import React, {useCallback, useEffect, useMemo, useState} from 'react';
 
 import {Input, Modal, Radio, Select} from 'antd';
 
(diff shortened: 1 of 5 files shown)


aabedraba:
Nice. 

When choosing multiple components to deploy, a namespace included, it enables the selector.

I think for now this should work, just in case someone is trying to deploy multiple resources at the same time; but maybe in the future we could make it smarter by figuring out if the other resources make use of the same namespace being created.

olensmar:
thanks @aabedraba for testing - we could detect this scenario - i.e. if
- the user is deploying multiple resources
- the set of resources contains a Namespace resource
- all other resources target that namespace
and in that scenario disable the namespace selector!?

How should we handle if the above is true except that not all resources target this namespace!? Perhaps an updated design of this interaction is needed - we could show a list of all resources with a selectable target namespace next to them - disabled for Namespace resources!? (this would be a separate issue)

aabedraba:
> we could show a list of all resources with a selectable target namespace next to them

I think this would be great, albeit only showing the list of resources without namespaces defined for them.

olensmar:
thanks @aabedraba - i've created https://github.com/kubeshop/monokle/issues/1376 for this - please add any more comments there!

src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
```diff
@@ -115,6 +116,10 @@ const ModalConfirmWithNamespaceSelect: React.FC<IProps> = props => {
     }
   }, [defaultOption, defaultNamespace, namespaces]);
 
+  useEffect(() => {
+    setDisableNamespaces(!resources.some(r => r.kind !== 'Namespace'));
```
devcatalin, line 120: Here it would be better to create a useMemo instead of state + useEffect:

```ts
const areNamespacesDisabled = useMemo(() => resources.some(r => r.kind === 'Namespace'), [resources]);
```

devcatalin (changes requested):

src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
olensmar: thanks @devcatalin - updated and added some additional logic to provide a (marginally) better UX

src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
```diff
@@ -115,6 +115,9 @@ const ModalConfirmWithNamespaceSelect: React.FC<IProps> = props => {
     }
   }, [defaultOption, defaultNamespace, namespaces]);
 
+  const onlyNamespaces = useMemo(() => !resources.some(r => r.kind !== 'Namespace'), [resources]);
```
devcatalin, line 118: This one is easier to read if you write it as:
`resources.every(r => r.kind === 'Namespace')`

devcatalin (changes requested):

src/components/molecules/ModalConfirmWithNamespaceSelect/ModalConfirmWithNamespaceSelect.tsx
olensmar: ha! i didn't know of `every` - thanks!

olensmar: @devcatalin fixed

devcatalin (approved):
Great!

Issue #1385, opened 2022-02-18
Improve syntax validation to warn for duplicate keys


olensmar:
We currently discard YAMLs that have (faulty) duplicate keys and do not follow the strict YAML 1.2 parsing rules - we should allow these and show any errors in the editor instead.

Reporting for visibility - as I am fixing this already...

Issue #1384, opened 2022-02-18
Syntax errors not highlighted in source editor in read-only mode


olensmar:
reporting for visibility; this needs to be configured explicitely - default is to show them only if editor is editable

Issue #1295, opened 2022-02-03
Find all values files in helm chart folder


devcatalin:
Right now Monokle only searches for values files that are in the same location of the `Chart.yaml` file.

We should search for values files in all subfolders of that helm chart folder.

PR #1388, opened 2022-02-19
Olelensmar/feat/values files in subfolders

olensmar:
This PR adds support for values files in subfolders to a helm chart. It also fixes / improves a bunch of other issues:
- Consistent parsing and improved validation of YAML errors
- Show YAML errors in Monaco in read-only mode
- Fixed deletion of Helm Charts
- Disabled "Add Resource" action for Helm / Values files in the File Explorer

## Changes

-

## Fixes

- #1295 
- #1385 
- #1384 

## How to test it

- Load a Helm Chart with values files in a subfolder to the chart itself - these should show up in the Helm pane also

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/redux/services/helm.ts
```diff
@@ -40,27 +46,53 @@ export function getHelmChartFromFileEntry(fileEntry: FileEntry, helmChartMap: He
...
 
 /**
- * check if the k8sResource is supported
- * @param resource
- * @returns @boolean
+ * check if the k8sResource is supported - currently excludes any files
+ * that seem to contain Helm template or Monokle Vanilla template content
  */
+
 export function isSupportedHelmResource(resource: K8sResource): boolean {
```
devcatalin, line 74: maybe we should rename this function to just `isSupportedResource` ?

src/redux/services/helm.ts
```diff
@@ -40,27 +46,53 @@ export function getHelmChartFromFileEntry(fileEntry: FileEntry, helmChartMap: He
...
+/**
+ * Adds the values file at the given path to the specified HelmChart
+ */
 
-  return Boolean(resource.text.match(helmVariableRegex)?.length) === false;
+export function createHelmValuesFile(fileEntry: FileEntry, helmChart: HelmChart, helmValuesMap: HelmValuesMapType) {
+  const helmValues: HelmValuesFile = {
+    id: uuidv4(),
+    filePath: fileEntry.filePath,
+    name: fileEntry.filePath.substring(path.dirname(helmChart.filePath).length + 1),
```
devcatalin, line 88: ins't `path.basename(fileEntry.filePath)` enough here?

src/redux/services/resource.ts
```diff
@@ -657,6 +658,10 @@ export function extractK8sResources(fileContent: string, relativePath: string) {
           documents[docIndex]
         );
       } else {
+        if (doc.warnings.length > 0) {
+          log.warn('Doc has warnings', doc.warnings, doc);
```
devcatalin, line 662: let's add `[extractK8sResources]:` at the beginning of this warning message

src/redux/services/validation.ts
```diff
@@ -54,6 +54,35 @@ export function validateResource(resource: K8sResource) {
...
+  const doc = parseDocument(resource.text);
+  if (doc.errors.length > 0) {
+    const lines = resource.text.split('\n');
+    if (lines[0] === '---') {
+      lines.shift();
+    }
+
+    errors.push(
+      ...doc.errors.map(err => {
+        const line = err.linePos ? lines[err.linePos[0].line].trim() : '';
```
devcatalin, line 69: is it possible for `err.linePos[0].line` to be undefined? maybe we should do a sanity check

devcatalin (changes requested):

src/redux/services/helm.ts
olensmar: no - this is meant to be relative to the containing HelmChart - so the name would be "subfolder/values.yaml" for a values file in a subfolder - not sure what else we could/should use as the name as just the filename might be confusing if the same name is used in different folders

devcatalin (approved):

PR #1399, opened 2022-02-22
feat: add nucleus.sh integration

erdkse:
This PR...

## Changes

- adds implementation of heap.io
- creates deviceID to define unique users

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


package.json
```diff
@@ -111,6 +111,7 @@
     "loglevel": "1.8.0",
     "luxon": "2.3.0",
     "micromatch": "4.0.4",
+    "mixpanel": "0.14.0",
```
devcatalin: do we still need this dependency?

src/components/organisms/TemplateModal/TemplateModal.tsx
```diff
@@ -51,6 +53,7 @@ const TemplateModal: React.FC<TemplateModalProps> = props => {
   const onClickSubmit = useCallback(
     (formDataList: Record<string, Primitive>[]) => {
       if (projectToCreate) {
+        trackEvent('START_FROM_A_TEMPLATE', {templateID: template.id});
```
devcatalin, line 56: should we track exactly which template was used or just the event name?

devcatalin (approved):
LGTM, but we should remove the API key as soon as possible and decide if we should save metadata along with the events or just the event names.
@olensmar @chlete what do you think?

package.json
chlete: I dont think so

src/components/organisms/TemplateModal/TemplateModal.tsx
chlete: @devcatalin both. In that way we can remove the least needed.
Actions are defined here: https://kubeshop.atlassian.net/wiki/spaces/KUBESHOP/pages/436043777/Telemetry#Actions

electron/utils.ts
```diff
@@ -142,3 +152,28 @@ export function askActionConfirmation({
 
   return choice === 0;
 }
+
+export const initNucleus = (isDev: boolean, app: any) => {
+  Nucleus.init('6218cf3ef5e5d2023724d89b', {
```
olensmar, line 157: @erdkse what id is this?

src/components/organisms/CreateProjectModal/CreateProjectModal.tsx
```diff
@@ -104,6 +105,7 @@ const CreateProjectModal: React.FC = () => {
       setFormStep(FormSteps.STEP_TWO);
     }
     if (!uiState.fromTemplate && values.rootFolder && values.name) {
+      trackEvent('CREATE_EMPTY_PROJECT');
```
olensmar, line 108: event ids should be in a eventconstants.ts file - in case we need to track the same event from multiple places

src/components/organisms/SettingsManager/SettingsManager.tsx
```diff
@@ -183,6 +201,26 @@ const SettingsManager: React.FC = () => {
               Show Cluster Selector
             </Checkbox>
           </S.Div>
+          <S.Div>
+            <S.Span style={{display: 'inline-block', marginRight: '8px'}}>Telemetry</S.Span>
+            <Tooltip title="DOCUMENTATION URL">
```
olensmar, line 206: should this link to something!?

electron/utils.ts
erdkse: This is my accounts credential - We will delete this when we move to Kubeshop's account

src/components/organisms/CreateProjectModal/CreateProjectModal.tsx
erdkse: Yeah - for sure I will create a separate PR to handle this.

src/components/organisms/SettingsManager/SettingsManager.tsx
erdkse: We talked to point to the privacy policy by this tooltip - but it is not ready yet.

chlete: We are preparing it with Lucky. He has already a template and we will decide how to do it

PR #1405, opened 2022-02-24
Now we can trigger contextmenu on file explorer section by right-clicking on tree items

egeren:
This PR...

## Changes

-

## Fixes

-

## How to test it
Right-click items that listed in file explorer.
-

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/ContextMenu/ContextMenu.tsx
+++ b/src/components/molecules/ContextMenu/ContextMenu.tsx
@@ -2,13 +2,14 @@ import {Dropdown} from 'antd';
 
 interface ContextMenuProps {
   overlay: React.ReactElement;
+  triggerOnRightClick?: boolean;
 }
 
 const ContextMenu: React.FC<ContextMenuProps> = props => {
-  const {overlay, children} = props;
+  const {overlay, children, triggerOnRightClick = false} = props;
 
   return (
-    <Dropdown overlay={overlay} trigger={['click']} placement="bottomCenter">
+    <Dropdown overlay={overlay} trigger={triggerOnRightClick ? ['contextMenu'] : ['click']} placement="bottomCenter">
       {children}
     </Dropdown>
   );
--- a/src/components/organisms/FileTreePane/TreeItem.tsx
+++ b/src/components/organisms/FileTreePane/TreeItem.tsx
@@ -250,41 +250,45 @@ const TreeItem: React.FC<TreeItemProps> = props => {
   );
 
   return (
-    <S.TreeTitleWrapper onMouseEnter={handleOnMouseEnter} onMouseLeave={handleOnMouseLeave}>
-      <S.TitleWrapper>
-        <S.TreeTitleText>{title}</S.TreeTitleText>
-        {canPreview(relativePath) && <EyeOutlined style={{color: isFileSelected ? Colors.blackPure : Colors.grey7}} />}
-      </S.TitleWrapper>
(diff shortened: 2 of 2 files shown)


erdkse:
Thank you for your PR @egeren ! LGTM! Maybe we can remove three dots, Wdyt folks @olensmar @devcatalin @andreiv1992 @DmitryEremenko ?

olensmar:
thanks @egeren - very nice!

@tommywhitehead what do yo think about removing the three dots - I still like them since they teach the user that there is a menu available.

Also - for consistency - I think we should add this to the resources in the navigator also - @devcatalin can you point @egeren to where that would be done? or should we handle it separately?

egeren:
> thanks @egeren - very nice!
> 
> @tommywhitehead what do yo think about removing the three dots - I still like them since they teach the user that there is a menu available.
> 
> Also - for consistency - I think we should add this to the resources in the navigator also - @devcatalin can you point @egeren to where that would be done? or should we handle it separately?

I'm glad to help! 
Today I will update this PR with that feature you mentioned and also in kustomize section.

src/components/molecules/SectionRenderer/ItemRenderer/ItemRenderer.tsx
```diff
@@ -70,81 +72,85 @@ function ItemRenderer<ItemType, ScopeType>(props: ItemRendererProps<ItemType, Sc
     scrollContainer.current?.scrollIntoView();
   }, [checkedResourceIds.length, itemInstance.shouldScrollIntoView]);
 
+  const Wrapper = ContextMenuWrapper.Component ? ContextMenuWrapper.Component : 'div';
```
devcatalin, line 75: Could this be written in a `useMemo` ? Maybe `ContextMenuWrapper.Component` could return `div` as default from useItemCustomization?

devcatalin (changes requested):

src/components/molecules/SectionRenderer/ItemRenderer/ItemRenderer.tsx
egeren: Actually I was thinking about that maybe some items must not have context menu option, so thats why I checked that and returned `div`. 
But, `KustomizationSectionBlueprint` is constant isn't it? So in every situtation that blueprint should have context menu (because every item has "Go to file" option in kustomize section). So I'm considering removing this line and rename `<Wrapper>` to `<ContextMenuWrapper.Component>`. What do you think about this one?

Sorry if I missed a point and please inform me about that.

devcatalin: Some SectionBlueprints could definitely not have a ContextMenu, like the sections related to Helm or the parent sections in the Navigator, I think we could leave it like this for now

julioramosest:
@egeren @olensmar UX input: I agree with leaving the 3 dots. Though it's duplicated CTA (click on 3 dots show same dropdown as right click) it is consistent with behaviour on Navigator column elements. Also, since we are not using tooltips to let the user know about the right click behaviour, keeping both is fine. We can rethink how to solve that in a simplified way later.

Ok also to dropdown, item separators on it, etc

julioramosest (approved):

erdkse:
@devcatalin Should we merge this?

devcatalin (approved):
Yes, let's merge it, lgtm

Issue #1287, opened 2022-02-03
Apply Changes not shown when editing/removing existing entries


olensmar:
For the Files: Exclude settings - it is only shown when adding a new exclusion

![image](https://user-images.githubusercontent.com/1917063/152318894-f43967e4-d1f5-4b27-b85d-86666f10eafe.png)

PR #1410, opened 2022-02-28
#1287 Apply Changes not shown

andreiv1992:
## Changes

-

## Fixes

- #1287

## How to test it

- Go to project settings
- add/edit/delete an option to include/ignore files
- after each change the apply changes should be visible

## Screenshots

- https://user-images.githubusercontent.com/5509112/156014901-b5f56c87-4d62-41f6-be11-a458d9ae75d1.mov
- 
## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/SettingsManager/Settings.tsx
+++ b/src/components/organisms/SettingsManager/Settings.tsx
@@ -13,6 +13,7 @@ import {
   K8S_VERSIONS,
   TOOLTIP_DELAY,
   TOOLTIP_K8S_SELECTION,
+  ROOT_FILE_ENTRY,
 } from '@constants/constants';
 import {
   AddExclusionPatternTooltip,
@@ -30,6 +31,8 @@ import {useAppDispatch, useAppSelector} from '@redux/hooks';
 import {updateShouldOptionalIgnoreUnsatisfiedRefs} from '@redux/reducers/main';
 import {isInClusterModeSelector} from '@redux/selectors';
 import {downloadSchema, schemaExists} from '@redux/services/k8sVersionService';
+import {setScanExcludesStatus, setScanIncludesStatus} from '@redux/reducers/appConfig';
+import {setRootFolder} from '@redux/thunks/setRootFolder';
 
 import FilePatternList from '@molecules/FilePatternList';
 
@@ -60,6 +63,8 @@ export const Settings = ({
 
   const resourceRefsProcessingOptions = useAppSelector(state => state.main.resourceRefsProcessingOptions);
   const uiState = useAppSelector(state => state.ui);
+  const appConfig = useAppSelector(state => state.config);
+  const fileMap = useAppSelector(state => state.main.fileMap);
 
   const isInClusterMode = useAppSelector(isInClusterModeSelector);
   const fileInput = useRef<HTMLInputElement>(null);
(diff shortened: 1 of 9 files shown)


src/components/organisms/SettingsManager/Settings.tsx
```diff
@@ -350,7 +360,11 @@ export const Settings = ({
           onChange={onChangeScanExcludes}
           tooltip={AddExclusionPatternTooltip}
           isSettingsOpened={isSettingsOpened}
-          type="excludes"
+          showApplyButton={appConfig.isScanExcludesUpdated === 'outdated'}
+          onApplyClick={() => {
+            dispatch(setScanExcludesStatus('applied'));
```
olensmar, line 365: does this have to be a separate reducer? or could the setRootFolder reducer set this property? (would apply above also)

andreiv1992: we should set includes and excludes in the `setRootFolder`, are we ok with this?

src/components/organisms/SettingsManager/Settings.tsx
```diff
@@ -60,6 +63,8 @@ export const Settings = ({
 
   const resourceRefsProcessingOptions = useAppSelector(state => state.main.resourceRefsProcessingOptions);
   const uiState = useAppSelector(state => state.ui);
+  const appConfig = useAppSelector(state => state.config);
+  const fileMap = useAppSelector(state => state.main.fileMap);
```
devcatalin, line 67: here I wouldn't select the entire `appConfig` and `fileMap` as it will re-render the component on any changes to their propertis, it's better to select directly only the state that you need

devcatalin (changes requested):

src/components/organisms/SettingsManager/Settings.tsx
andreiv1992: Done

src/components/organisms/SettingsManager/Settings.tsx
andreiv1992: Thanks for the suggestions, it looks a lot more cleaner now

src/components/organisms/SettingsManager/Settings.tsx
```diff
@@ -63,8 +63,8 @@ export const Settings = ({
 
   const resourceRefsProcessingOptions = useAppSelector(state => state.main.resourceRefsProcessingOptions);
   const uiState = useAppSelector(state => state.ui);
-  const appConfig = useAppSelector(state => state.config);
-  const fileMap = useAppSelector(state => state.main.fileMap);
+  const {isScanIncludesUpdated, isScanExcludesUpdated} = useAppSelector(state => state.config);
+  const filePath = useAppSelector(state => state.main.fileMap[ROOT_FILE_ENTRY].filePath);
```
devcatalin, line 66: could this throw an error if you open the Settings drawer without having a project open?
better to write it as: `state.main.fileMap[ROOT_FILE_ENTRY]?.filePath`

devcatalin (changes requested):

devcatalin:
Looks good to me know! Only a really small thing that I left a comment with

src/components/organisms/SettingsManager/Settings.tsx
andreiv1992: tested it and it does not but it does not hurt to have this in there

devcatalin (approved):

PR #1431, opened 2022-03-04
chore: Telemetry reviews fixes

erdkse:
This PR...

## Changes

- add event names into constant

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/utils/shell.ts
+++ b/src/utils/shell.ts
@@ -37,12 +37,12 @@ export function openUrlInExternalBrowser(url?: string) {
 }
 
 export function openDocumentation() {
-  trackEvent(WINDOW_HELP_LINK, {linkID: 'documentation', osType: os.type});
+  trackEvent(WINDOW_HELP_LINK, {linkID: 'documentation'});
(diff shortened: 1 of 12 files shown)


erdkse:
Created an issue to make EventErrors into readable format https://github.com/sindresorhus/electron-unhandled/issues/26

src/utils/shell.ts
```diff
@@ -37,12 +37,12 @@ export function openUrlInExternalBrowser(url?: string) {
 }
 
 export function openDocumentation() {
-  trackEvent('WINDOW_HELP_LINK', {linkID: 'documentation', osType: os.type});
+  trackEvent(WINDOW_HELP_LINK, {linkID: 'documentation', osType: os.type});
```
olensmar, line 40: do we really need to add os.type to this event? don't we get it automatically through nucleus?

erdkse: yeah we have - There is no need to keep this value You are right

olensmar (approved):

PR #1432, opened 2022-03-04
feat: Improve plugins install URL support

erdkse:
This PR...

## Changes

- Drop need of repository field in the plugin template, Now we use repository URL instead since we know where we install from
- Detecting branch and installing into specifically created branch folder
- Checking all JSON files if they are not valid don't let the plugin to be installed
- Show better error messages
- Some bug fixes

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
@erdkse which improvements have you added? (so we know what to test/expect)

erdkse:
I will add my conversation with Catalin here as a reference @olensmar 

Screen Shot 2022-03-04 at 10 35 50

olensmar:
thanks @erdkse!

src/components/organisms/PluginManager/PluginInstallModal.tsx
```diff
@@ -85,12 +138,22 @@ function PluginInstallModal(props: {isVisible: boolean; onClose: () => void}) {
...
-        <div>
-          <p>{errorMessage}</p>
-        </div>
-      )}
+      <Form layout="vertical" form={pluginForm} onFinish={onFinish} initialValues={() => formValues}>
+        <Form.Item
+          name="pluginUrl"
+          label="URL"
+          required
+          tooltip="pluginUrl"
```
olensmar, line 146: perhaps a better tooltip? Should we add some text to the dialog showing which types of URLs we support?

erdkse: Ahh yeah, I forgot that thank you for noticing - The user can't enter an invalid URL right now We show an error if it happens so I am not sure If we need to show which types of URLs we support - besides If they enter a valid git URL we only download files in the default main branch - Do you see value to support also other branches rather than the default main one? It is easy to add

olensmar: i think we should add branch support if it is easy - and some kind of text/placeholder to the form to indicate what URL(s) we support

src/constants/tooltips.ts
```diff
@@ -60,3 +60,4 @@ export const NewResourceTooltip = `Create new resource (${KEY_CTRL_CMD} + N)`;
 export const NewPreviewConfigurationTooltip = 'Create a new Preview Configuration';
 export const DisableEventTrackingTooltip = 'Disables collection of usage metrics';
 export const DisableErrorReportingTooltip = 'Disables collection of error reports';
+export const PLUGIN_URL_INPUT_TOOLTIP = 'Enter the repository/branch url of the plugin you want to add!';
```
olensmar, line 63: please rename constant to PluginUrlInputTooltip for consistency..

src/components/organisms/PluginManager/PluginInstallModal.tsx
```diff
@@ -85,12 +128,22 @@ function PluginInstallModal(props: {isVisible: boolean; onClose: () => void}) {
...
+          tooltip={PLUGIN_URL_INPUT_TOOLTIP}
+          rules={[
+            {
+              required: true,
+              message: 'Please provide your plugin URL!',
+            },
+          ]}
+        >
+          <Input ref={inputRef} placeholder="Enter Plugin URL" />
+        </Form.Item>
```
olensmar, line 145: I still think we should have some kind of info on what kind of url(s) we support... @chlete what do you think?

olensmar (approved):
LGTM!

PR #1434, opened 2022-03-04
Apply on ubuntu

olensmar:
This PR 
- moves all apply/preview actions to the main thread and introduces a common mechanism for invoking commands from the renderer thread -> this fixes apply on ubuntu as all commands now are run with the correct environment
- removes the PROCESS_ENV constant; in the main thread we should just use process.env, in the renderer thread we should use getMainThreadEnv()
- updates electron and kube-client dependencies
- improves the deploy button to show "Install" for values files 

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test


andreiv1992 (approved):

src/redux/thunks/applyResource.ts
```diff
@@ -53,43 +52,18 @@ function applyK8sResource(
 function applyKustomization(
   resource: K8sResource,
   fileMap: FileMapType,
-  kubeconfig: string,
   context: string,
-  kustomizeCommand: KustomizeCommandType,
+  projectConfig: ProjectConfig,
```
devcatalin, line 56: do we need the entire `projectConfig` here? 
I'm asking because using a selector to get the entire `projectConfig` in a component will make that component re-render each time any property from the config has changed. And from what I saw it's selected in a couple different components

devcatalin, line 56: do we need the entire `projectConfig` here? 
I'm asking because using a selector to get the entire `projectConfig` in a component will make that component re-render each time any property from the config has changed. And from what I saw it's selected in a couple different components

devcatalin:
Other than the comment I left above, everything else looks good to me.
Will you make the changes needed for running the command for preview configurations as well?
The PR is merged to main so you can pull it in this branch

devcatalin:
Other than the comment I left above, everything else looks good to me.
Will you make the changes needed for running the command for preview configurations as well?
The PR is merged to main so you can pull it in this branch

olensmar:
> Other than the comment I left above, everything else looks good to me. Will you make the changes needed for running the command for preview configurations as well? The PR is merged to main so you can pull it in this branch

changes made - seems to work fine!

src/redux/thunks/applyResource.ts
olensmar: this method also uses projectConfig.kubeConfig?.path so it seemed better to pass the config instead of individual args

devcatalin (approved):

PR #1443, opened 2022-03-07
moved reducers into thunks with related refactorings

olensmar:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/kindhandlers/common/customMatchers.ts
```diff
@@ -4,7 +4,7 @@ import {K8sResource} from '@models/k8sresource';
  * Matcher that ensures the source and target namespace are the same
  */
 
-export function implicitNamespaceMatcher(sourceResource: K8sResource, targetResource: K8sResource, value: string) {
+export function implicitNamespaceMatcher(sourceResource: K8sResource, targetResource: K8sResource, _: string) {
```
andreiv1992, line 7: Should _ still be here since it's not used?

src/redux/services/fileEntry.ts
```diff
@@ -430,7 +429,7 @@ export function reloadFile(
   if (isHelmChartFile(absolutePath)) {
     reloadHelmChartFile(fileEntry, state.fileMap, state.helmChartMap);
   } else if (shouldReloadResourcesFromFile(fileEntry, projectConfig)) {
-    reloadResourcesFromFileEntry(fileEntry, state, schemaVersion, userDataDir);
+    reloadResourcesFromFileEntry(fileEntry, state, projectConfig.k8sVersion || PREDEFINED_K8S_VERSION, userDataDir);
```
andreiv1992, line 432: I think it would be a good idea to create a function which returns the value of the k8sVersion. I see the same code bellow as well `projectConfig.k8sVersion || PREDEFINED_K8S_VERSION`

src/redux/services/resource.ts
```diff
@@ -496,7 +496,7 @@ export function reprocessResources(
  * Establishes refs for all resources in specified resourceMap
  */
 
-export function processParsedResources(
+export function processResources(
```
andreiv1992, line 499: It's a bit hard to keep track of functions which have more than 2-3 params, it would be a good idea to send a single params as an object with each function param as a property(you know what goes where and it's less prone to human error in sending wrong params).
I don't know if this falls under this refactoring

src/redux/services/resourceRefs.ts
```diff
@@ -694,3 +701,28 @@ export function findResourcesToReprocess(resource: K8sResource, resourceMap: Res
 
   return resources;
 }
+
+/**
+ * Reprocess all optoinal refs for the specified resources
+ */
+
+export function reprocessOptionalRefs(
```
andreiv1992, line 709: Same as before it's better to have an object and send properties there than sending multiple arguments to a function

src/redux/thunks/previewHelmValuesFile.ts
```diff
@@ -68,8 +66,8 @@ export const previewHelmValuesFile = createAsyncThunk<
 
       if (result.stdout) {
         return createPreviewResult(
-          String(k8sVersion),
-          String(userDataDir),
+          projectConfig.k8sVersion || PREDEFINED_K8S_VERSION,
```
andreiv1992, line 69: The `projectConfig.k8sVersion || PREDEFINED_K8S_VERSION` is used once again. If we will want to refactor this it will be easier to have a single point of entry

src/redux/thunks/previewKustomization.ts
```diff
@@ -37,10 +37,11 @@ export const previewKustomization = createAsyncThunk<
   }
 >('main/previewKustomization', async (resourceId, thunkAPI) => {
   const state = thunkAPI.getState().main;
-  const k8sVersion = thunkAPI.getState().config.projectConfig?.k8sVersion;
-  const userDataDir = thunkAPI.getState().config.userDataDir;
   const projectConfig = currentConfigSelector(thunkAPI.getState());
+  const k8sVersion = projectConfig.k8sVersion || PREDEFINED_K8S_VERSION;
```
andreiv1992, line 41: again `projectConfig.k8sVersion || PREDEFINED_K8S_VERSION`

src/redux/thunks/setRootFolder.ts
```diff
@@ -66,14 +68,14 @@ export const setRootFolder = createAsyncThunk<
   rootEntry.children = files;
 
   processKustomizations(resourceMap, fileMap);
-  processParsedResources(
-    String(projectConfig?.k8sVersion),
+  processResources(
+    projectConfig.k8sVersion || PREDEFINED_K8S_VERSION,
```
andreiv1992, line 72: again `projectConfig.k8sVersion || PREDEFINED_K8S_VERSION`

src/redux/thunks/updateFileEntry.ts
```diff
@@ -0,0 +1,96 @@
...
+  'main/updateFileEntry',
+  async (payload: UpdateFileEntryPayload, thunkAPI: {getState: Function; dispatch: Function}) => {
+    const state: RootState = thunkAPI.getState();
+    const projectConfig = currentConfigSelector(state);
+    const schemaVersion = projectConfig.k8sVersion || PREDEFINED_K8S_VERSION;
+    const userDataDir = String(state.config.userDataDir);
+
+    try {
+      const fileEntry = state.main.fileMap[payload.path];
+      if (fileEntry) {
```
andreiv1992, line 29: I think that the logic from this file would be made a bit simple if we can exit early
`
if (!fileEntry) {log...}
`
and so of if it's possible

src/redux/services/fileEntry.ts
olensmar: you're right - will fix!

src/redux/services/resource.ts
olensmar: yeah - sounds like a general refactoring we should do - i can create an issue

src/kindhandlers/common/customMatchers.ts
olensmar: yes my bad - mixing up languages - changing back to value

olensmar: @andreiv1992 the value is included in the contract for matcher methods - its just that this matcher doesn't use it

devcatalin (approved):
LGTM

PR #1471, opened 2022-03-15
feat: Release Notes modal and drawer

devcatalin:
This PR...

## Changes

-

## Fixes

-

## How to test it

- The modal should pop-up when opening Monokle for the first time or when there is a new version update
- The drawer can be opened from the question mark icon from the top right by clicking on the "New in {version}" button

## Screenshots

image

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/PageHeader/HelpMenu.tsx
```diff
@@ -43,6 +60,18 @@ const HelpMenu = () => {
         </Row>
       </MenuItem>
 
+      <MenuItem onClick={onClickReleaseNotes} key="releasenotes" style={{paddingRight: '20px'}}>
```
andreiv1992, line 63: I can see that this was not introduced by you, it seems that all the menu items have a right padding of 20px, maybe we can move this to a styled file?

src/components/organisms/ReleaseNotes/ReleaseNotes.tsx
```diff
@@ -0,0 +1,88 @@
+import {useEffect, useState} from 'react';
```
andreiv1992, line 1: there are some inline styles, can we create a styled file

andreiv1992 (changes requested):

andreiv1992 (approved):

PR #1545, opened 2022-03-31
fix: updated electron and inlined fix-path dependency to avoid lockup

olensmar:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


electron/app/commands.ts
```diff
@@ -136,11 +135,6 @@ export const runCommand = (options: CommandOptions, event: Electron.IpcMainEvent
 
   const result: CommandResult = {exitCode: null, signal: null};
 
-  fixPath();
```
olensmar, line 139: removed it here since it is called when creating the window

devcatalin: I was thinking of adding it to the `electron/main.ts` before everything else loads just to make sure all/any future features that we'll add and might depend on the PATH variable will work well

olensmar: thanks @devcatalin I will move it there - how can I test that it works?

devcatalin (approved):

PR #1547, opened 2022-04-01
fix: allow to create multiple resources from template

erdkse:
This PR...

## Changes

-

## Fixes

- It allows to create multiple resources from template

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/redux/services/templates.ts
```diff
@@ -128,24 +130,33 @@ export const createUnsavedResourcesFromVanillaTemplate = async (
...
-        createdResources.push(resource);
-      });
+      objects = [...objects, ...extractObjectsFromYaml(resourceText)];
     });
+
+  const createdResources: K8sResource[] = [];
+
+  for (const obj of objects) {
+    // eslint-disable-next-line no-await-in-loop
+    await sleep(500);
```
andreiv1992, line 146: Why do we have a timeout here?

erdkse: Because The state value gets overwritten if we call 2`createResource` functions immediately.

devcatalin: I'd like to test this for some time to see if it can be fixed in another way.
I think even a new action that creates multiple unsaved resources at once could be better... we should avoid relying on timeouts like this one

src/redux/services/templates.ts
```diff
@@ -128,24 +129,24 @@ export const createUnsavedResourcesFromVanillaTemplate = async (
       }
     }
   );
-  const createdResources: K8sResource[] = [];
+
+  let objects: any[] = [];
```
devcatalin, line 133: there is a `PossibleResource` type in the `redux/services/manifest-utils` file, could you use that here?
maybe move it to the `models/appstate` first and then import it in both places.

also add something like this to be sure the properties exist:
```
obj &&
typeof obj.apiVersion === 'string' &&
typeof obj.kind === 'string' &&
typeof obj.metadata?.name === 'string'
```

You could probably extrat that into a type guard named `isPossibleResource` and have it in the `appstate` as well

src/utils/index.ts
```diff
@@ -8,3 +8,6 @@ export const generateId = (prefix: string = 'id'): string => {
 export const uniqueArr = <T>(arr: Array<T>): Array<T> => {
   return Array.from(new Set(arr));
 };
+
+// eslint-disable-next-line no-promise-executor-return
+export const sleep = (ms: number) => new Promise(res => setTimeout(() => res(null), ms));
```
devcatalin, line 13: do we still need this or can it be removed?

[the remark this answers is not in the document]
erdkse: I kept it in case of if we use, but you are right removing it

src/redux/services/unsavedResource.ts
```diff
@@ -66,3 +66,51 @@ export function createUnsavedResource(
...
+        },
+      };
+      resourceMap[resourceID].text = stringify(resourceMap[resourceID].content);
+    } else {
+      resourceMap[resourceID].text = createDefaultResourceText(input);
+      resourceMap[resourceID].content = parseYamlDocument(resourceMap[resourceID].text).toJS();
+    }
+  });
+
+  const newResources: K8sResource[] = Object.keys(resourceMap).map(resourceID => ({
```
devcatalin, line 99: let's rename `resourceID` to `resourceId` to be consistent with other places where we use it

devcatalin (changes requested):

src/redux/services/templates.ts
erdkse: where do you think to use this check? Do we need it really?

devcatalin (approved):
LGTM

PR #1555, opened 2022-04-04
Fix black screen due on selecting an existing folder

WitoDelnat:
This PR fixes a minor bug which prevents me from using Monokle. After selecting an existing folder, I used the argo-cd/manifests as suggested, the screen goes black completely. 

The interaction causing the issue is that I previously tried Monokle and had put the file explorer on hidden. Now after loading a new project it breaks because of a division by zero on this line: src/components/organisms/PaneManager/PaneMangerSplitView.tsx:91.

## Fixes

- Fix black screen on project selection.

## How to test it

- Hide the file explorer
- Open a new project
- The screen should no longer be black

## Screenshots

- Screen capture of issue

https://user-images.githubusercontent.com/7761005/161520253-ba7769d6-593b-4964-a530-4de77af6a175.mov

## Checklist

- [x ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/PaneManager/PaneManagerSplitView.tsx
```diff
@@ -88,7 +88,12 @@ const PaneManagerSplitView: React.FC = () => {
     }
 
     if (!leftActive) {
-      dispatch(setPaneConfiguration({...paneConfiguration, navWidth: navPaneWidth / splitViewContainerWidth}));
+      dispatch(
+        setPaneConfiguration({
+          ...paneConfiguration,
+          navWidth: splitViewContainerWidth === 0 ? 0 : navPaneWidth / splitViewContainerWidth,
```
WitoDelnat, line 94: Still trying to figure out if `0` is the correct value or whether it needs to be simply current navPaneWidth. I'll try playing around with Monokle to see what looks best.

WitoDelnat: It used to default to navPaneWidth and this problem was likely introduced in fd52da8faf2b5a6d4e14c13ac15f233e2c4abc0c. I'm reverting it to the code there that handled this null division.

devcatalin (approved):
Good catch, LGTM!

PR #1557, opened 2022-04-04
feat: render custom icons for plugins/templates from url and path png…

jmelnych:
This PR...
 render custom icons for plugins/templates from URL and/or path for SVG and PNG

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots
Screenshot 2022-04-04 at 12 15 08

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/atoms/ImageIcon/ImageIcon.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+type ImageProps = {
+  src?: string;
+  width?: number;
+  height?: number;
+  altText: string;
+  fallback: string;
+};
+
+const Image = styled.img<{height?: number; width?: number}>`
```
erdkse, line 11: let's divide style codes into a different file. We use this pattern.

src/components/atoms/ImageIcon/ImageIcon.tsx
```diff
@@ -0,0 +1,27 @@
...
+  width: ${props => (props.width ? `${props.width}px` : '32px')}};
+  `;
+
+const ImageIcon = ({src = '', width, height, altText, fallback}: ImageProps) => {
+  let source;
+  if (src.includes('http')) {
+    source = src;
+  } else if (src) {
+    source = 'file://'.concat(src);
+  }
```
devcatalin, line 22:
```suggestion
  const source = useMemo(() => {
    if (src.includes('http')) {
      return src;
    } else if (src) {
      return 'file://'.concat(src);
    }
  }, [src]);
```

devcatalin (changes requested):

Issue #1495, opened 2022-03-22
Show error/warning indicators in collapsed view


olensmar:
When collapsing sections in the navigator that contain resources with warnings/errors I would like to see some kind of indicator next to each section that contains errors/warnings to I know where to find those resources... 

<img width="523" alt="image" src="https://user-images.githubusercontent.com/1917063/159488529-5a8b5bc9-89d3-477b-9772-25fa14836ed4.png">

PR #1569, opened 2022-04-06
Aggregate warnings on collapse nav section

WitoDelnat:
This PR closes https://github.com/kubeshop/monokle/issues/1495

## Changes

- Aggregates warnings on collapsing a navigation section.

## How to test it

- Open a project with warnings or errors
- Collapse a nav section with said warnings
- Validate that warnings are aggregated.

## Screenshots

https://user-images.githubusercontent.com/7761005/161929246-c4be3314-8a17-4e89-945c-5daf87d1d961.mov

## Checklist

- [x ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/SectionRenderer/SectionHeader.tsx
+++ b/src/components/molecules/SectionRenderer/SectionHeader.tsx
@@ -1,9 +1,10 @@
 import {useCallback, useMemo, useState} from 'react';
 
-import {SectionBlueprint, SectionInstance} from '@models/navigator';
+import {SectionBlueprint, SectionCustomComponent, SectionInstance} from '@models/navigator';
 
 import {useAppDispatch} from '@redux/hooks';
 
+import SectionHeaderDefaultNameCounter from './SectionHeaderDefaultNameCounter';
 import {useSectionCustomization} from './useSectionCustomization';
 
 import * as S from './styled';
@@ -39,6 +40,11 @@ function SectionHeader(props: SectionHeaderProps) {
     sectionBlueprint.customization
   );
 
+  const Counter: SectionCustomComponent = useMemo(
+    () => NameCounter.Component ?? SectionHeaderDefaultNameCounter,
+    [NameCounter]
+  );
+
   const toggleCollapse = useCallback(() => {
     if (isCollapsed) {
       expandSection();
@@ -46,20 +52,6 @@ function SectionHeader(props: SectionHeaderProps) {
       collapseSection();
     }
   }, [isCollapsed, expandSection, collapseSection]);
-
-  const counter = useMemo(() => {
-    const counterDisplayMode = sectionBlueprint.customization?.counterDisplayMode;
-    if (!counterDisplayMode || counterDisplayMode === 'descendants') {
-      return sectionInstance?.visibleDescendantItemIds?.length || 0;
-    }
-    if (counterDisplayMode === 'items') {
-      return sectionInstance?.visibleItemIds.length;
-    }
-    if (counterDisplayMode === 'subsections') {
-      return sectionInstance?.visibleChildSectionIds?.length || 0;
-    }
-    return undefined;
-  }, [sectionInstance, sectionBlueprint]);
 
   const onCheck = useCallback(() => {
     if (!sectionInstance.checkable || !sectionInstance.visibleDescendantItemIds) {
@@ -134,13 +126,7 @@ function SectionHeader(props: SectionHeaderProps) {
               {name}
             </S.Name>
 
-            {NameCounter.Component ? (
-              <NameCounter.Component sectionInstance={sectionInstance} onClick={toggleCollapse} />
-            ) : (
-              counter !== undefined && (
-                <S.Counter selected={sectionInstance.isSelected && isCollapsed}>{counter}</S.Counter>
-              )
-            )}
+            <Counter sectionInstance={sectionInstance} sectionBlueprint={sectionBlueprint} onClick={toggleCollapse} />
 
             <S.BlankSpace level={level} onClick={toggleCollapse} />
 
--- a/src/components/molecules/SectionRenderer/SectionHeader.tsx
+++ b/src/components/molecules/SectionRenderer/SectionHeader.tsx
@@ -35,7 +35,9 @@ function SectionHeader(props: SectionHeaderProps) {
   const dispatch = useAppDispatch();
   const [isHovered, setIsHovered] = useState<boolean>(false);
 
-  const {NameDisplay, NamePrefix, NameSuffix, NameContext} = useSectionCustomization(sectionBlueprint.customization);
+  const {NameDisplay, NamePrefix, NameSuffix, NameContext, NameCounter} = useSectionCustomization(
+    sectionBlueprint.customization
+  );
 
   const toggleCollapse = useCallback(() => {
     if (isCollapsed) {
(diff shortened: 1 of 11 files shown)


WitoDelnat:
Julio is going to give me an updated icon as contrast is not very good, but functionally everything is there.

Screenshot 2022-04-05 at 17 36 03

src/navsections/K8sResourceSectionBlueprint/ResourceKindSectionNameCounter.tsx
```diff
@@ -0,0 +1,43 @@
+import React, {useMemo} from 'react';
+
+import {SectionCustomComponentProps} from '@models/navigator';
+
+import {useAppSelector} from '@redux/hooks';
+
+import {countResourceErrors, countResourceWarnings} from '@utils/resources';
+
+import * as S from '../../components/molecules/SectionRenderer/styled';
+import * as WarningStyle from '../../components/organisms/NavigatorPane/WarningAndErrorsDisplay.styled';
```
WitoDelnat, line 10: I do not like this too much. Suggestions on how you share style within this project?

src/components/molecules/SectionRenderer/SectionHeader.tsx
```diff
@@ -35,7 +36,14 @@ function SectionHeader(props: SectionHeaderProps) {
   const dispatch = useAppDispatch();
   const [isHovered, setIsHovered] = useState<boolean>(false);
 
-  const {NameDisplay, NamePrefix, NameSuffix, NameContext} = useSectionCustomization(sectionBlueprint.customization);
+  const {NameDisplay, NamePrefix, NameSuffix, NameContext, NameCounter} = useSectionCustomization(
+    sectionBlueprint.customization
+  );
+
+  const Counter: SectionCustomComponent = useMemo(
+    () => NameCounter.Component ?? SectionHeaderDefaultNameCounter,
```
WitoDelnat, line 44: Curious to what you think about this @devcatalin. I can always revert this and do it conditionally as you can see in https://github.com/kubeshop/monokle/pull/1569/commits/d467cc016af1b5d1e283a7fe185a1b3af4b12534.

src/navsections/K8sResourceSectionBlueprint/ResourceKindSectionNameCounter.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+function ResourceKindSectionCounter({sectionInstance, onClick}: SectionCustomComponentProps) {
+  const {id, isSelected, itemIds} = sectionInstance;
+  const isCollapsed = useAppSelector(state => state.navigator.collapsedSectionIds.includes(id));
+  const resourceMap = useAppSelector(state => state.main.resourceMap);
+  const resources = useMemo(() => itemIds.map(itemId => resourceMap[itemId]), [itemIds, resourceMap]);
```
WitoDelnat, line 16: I assume an invariant of this codebase is that all resources can be found in the resource map, otherwise this might give some `undefined` values which I have to filter.

src/navsections/K8sResourceSectionBlueprint/ResourceKindSectionNameCounter.tsx
devcatalin: I would say just duplicate the styles that you need for now. Or you could create a `components` folder under `src/styles` and move the reusable styles there.

I think so far in the code-base we duplicated the needed styles just to be able to change their styles separately if we did big redesigns.

devcatalin: Or, your idea to extract these as Atoms would be good as well, but only if they have some kind of logic.. if the component is strictly a styled component with no logic, maybe having them in the `styles` folder is better?!

Any thoughts @andreiv1992 @erdkse @jmelnych ?

src/navsections/K8sResourceSectionBlueprint/ResourceKindSectionNameCounter.tsx
devcatalin: It would be safer to filter out undefined values

src/components/molecules/SectionRenderer/SectionHeader.tsx
devcatalin: This looks good!

src/models/navigator.ts
```diff
@@ -49,6 +49,7 @@ export interface ItemCustomization {
 
 export type SectionCustomComponentProps = {
   sectionInstance: SectionInstance;
+  sectionBlueprint?: SectionBlueprint<any>;
```
devcatalin, line 52: Let's remove this from here because each `sectionInstance` has the same `id` as it's `sectionBlueprint`.
Thus, in your custom component you can import the `sectionBlueprintMap` and do the following to get the blueprint:
```const sectionBlueprint = sectionBlueprintMap.getById(sectionInstance.id);```

File: src/navsections/K8sResourceSectionBlueprint/ResourceKindSectionNameCounter.tsx (hunk 3 above)
Comment (devcatalin) [inline src/navsections/K8sResourceSectionBlueprint/ResourceKindSectionNameCounter.tsx:16]: I also suggest having a single `useAppSelector` to get the resources if you don't need the `resourceMap` individually.
```suggestion
  const resources = useAppSelector(state => itemIds.map(itemId => state.main.resourceMap[itemId]).filter((r): r is K8sResource => r !== undefined));
```

devcatalin (changes requested):

src/navsections/K8sResourceSectionBlueprint/ResourceKindSectionNameCounter.tsx
jmelnych: As for sharable styles - we can have all app shared styles in App.styled.ts file, or create a folder like @devcatalin suggested. As for components atoms - I think that components imply to have some render logic. If it is solely about styles - let's keep that in styles, this is my suggestion.

src/models/navigator.ts
WitoDelnat: All parent components have the blue print injected so it's a bit odd to rely on a global variable in the leaf component. My feeling is that injecting dependencies is easier to reason about than using globals. I'm curious to your reasoning behind using the map? (Do note that I do not mind changing it, trying to understand for the future)

devcatalin: The blueprint object is not serializable and it probably won't work well when React checks if the blueprint sent as prop as changed. It could lead to unnecessary re-renders.

I don't see any problem in relying on global objects when it's something used in many places of the app. This is similar to the `store` object from Redux as which you could import in any file even if it's not a React component and you can still read the state or dispatch actions to it.

devcatalin: And yes, it should be exceptional when those components need the blueprint.

We have `getMeta` on items if we need to create `any` object we want on the `instance.meta`

WitoDelnat: That makes sense, I'm glad I asked as I would not have guessed that. Thanks for the the clarification.

src/navsections/K8sResourceSectionBlueprint/ResourceKindSectionNameCounter.tsx
WitoDelnat: I decided to go for duplicated style, neither extraction felt right.
1/ The icon in navigator does not have the "selected" logic.
2/ The icon in a single resource does not include a count.

The current look is this

Screenshot 2022-04-07 at 10 04 07

devcatalin: Looks good to me!

devcatalin (approved):
LGTM!

PR #1571, opened 2022-04-06
Fix resource validation

WitoDelnat:
This PR fixes resource error validation

Note: this does not add strict schema validation, for instance `fdsfasd: "foo"`. This fixes purely the normal type validation.

## Fixes

- resource validation

## How to test it

- See video

## Screenshots

This video shows how the error shows after the refactor and when stashed (before) it did not show any error.

https://user-images.githubusercontent.com/7761005/161949440-557adcfb-5119-45f3-be9b-67fd76b3f12b.mov

## Checklist

- [ x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/redux/services/validation.ts
```diff
@@ -49,149 +47,151 @@ export function validateResource(resource: K8sResource, schemaVersion: string, u
...
+  schemaVersion: string,
+  userDataDir: string
+): Ajv.ValidateFunction | undefined {
+  const resourceSchema = getResourceSchema(resource, schemaVersion, userDataDir);
 
-        resource.validation = {
-          isValid: errors.length === 0,
-          errors: errors && errors.length > 0 ? errors : [],
-        };
-      } catch (e) {
```
WitoDelnat, line 160: This try-catch should be unnecessary as seen from the AJV docs. Please double check if I'm overlooking why this should throw.

> Compiled parsers do NOT throw exceptions
>
> Compiled parsers, unlike JSON.parse, do not throw the exception in case JSON string is not a valid JSON or in case data is invalid according to the schema. As soon as the parser determines that either JSON or data is invalid, it returns undefined and reports error and position via parsers properties message and position.

devcatalin: Thanks @WitoDelnat , this is indeed the case so the try catch is not needed

devcatalin:
The code looks good to me so far, nice work!
Let's set the ajv validation to be strict so it shows errors on the properties that are added in the yaml but are not present in the schema. This will be useful for users when they change between k8s schemas to catch properties that are not needed.

Also, I noticed that the highlighting of errors in the Monaco editor do no show up:

image

Here the line with the `replicas` should have a red underline. After enabling strict mode, the line with the `test` property should be underlined as well.

WitoDelnat:
@devcatalin strict schema checking is not trivial since AJV and Monaco work independently. Setting AJV on strict will give desync. The common denominator is the JSON schema. In Monokle 1.5.0 this worked because the json schema contained `additionalProperties` everywhere, while more recent JSON schemas have removed this.

devcatalin (approved):
Great work!

Issue #1590, opened 2022-04-11
Files get automatically collapsed


devcatalin:
If you expand the folders in the FileTree, then go to the Helm tab and back to the Files tab, the files will be collapsed entirely.

**Expected behavior:**
The FileTree should maintain it's collapsed/expanded state between tab changes.

PR #1596, opened 2022-04-12
fix: add minor file explorer improvements

WitoDelnat:
This PR closes https://github.com/kubeshop/monokle/issues/1590 and fixes a few issues with file explorer.

⚠️ There is still a minor issue where parents are expanded on switching views - I think it's non-blocking and we can continue as-is while upstream fixes it later on.

## Changes

- Switching views (e.g. File explorer to Helm and back) maintains expanded folders.
- Improve label of toggling collapse/expand. 

## Fixes

- Reloading maintains expanded folders.
- Scroll to selected on mount.

## How to test it

- Open a project and expand some folders.
- Switch to Helm and switch back.
- Validate that the folders are still open.

## Screenshots

https://user-images.githubusercontent.com/7761005/162939561-1625c625-e3df-453c-b4f7-63054c329347.mov

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/FileTreePane/FileTreePane.tsx
+++ b/src/components/organisms/FileTreePane/FileTreePane.tsx
@@ -198,14 +198,14 @@ const FileTreePane = () => {
 
   function highlightFilePath(filePath: string) {
     const paths = filePath.split(path.sep);
-    const keys: Array<React.Key> = [];
+    const keys: Array<React.Key> = [ROOT_FILE_ENTRY];
 
     for (let c = 1; c < paths.length; c += 1) {
       keys.push(paths.slice(0, c + 1).join(path.sep));
     }
 
     let node: TreeNode | undefined = tree || undefined;
-    for (let c = 0; c < keys.length && node; c += 1) {
+    for (let c = 1; c < keys.length && node; c += 1) {
       node = node.children.find((i: any) => i.key === keys[c]);
     }
 
--- a/src/components/organisms/FileTreePane/FileTreePane.tsx
+++ b/src/components/organisms/FileTreePane/FileTreePane.tsx
@@ -12,7 +12,7 @@ import micromatch from 'micromatch';
 import path from 'path';
 
 import {FILE_TREE_HEIGHT_OFFSET, ROOT_FILE_ENTRY, TOOLTIP_DELAY} from '@constants/constants';
-import {FileExplorerChanged, ReloadFolderTooltip, ToggleTreeTooltip} from '@constants/tooltips';
+import {CollapseTreeTooltip, ExpandTreeTooltip, FileExplorerChanged, ReloadFolderTooltip} from '@constants/tooltips';
 
 import {AlertEnum} from '@models/alert';
 import {FileMapType, ResourceMapType} from '@models/appstate';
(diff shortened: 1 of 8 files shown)


src/components/organisms/FileTreePane/TreeItem.tsx
```diff
@@ -72,11 +72,11 @@ const TreeItem: React.FC<TreeItemProps> = props => {
 
   const getBasename = osPlatform === 'win32' ? path.win32.basename : path.basename;
 
-  const isRoot = fileMap[ROOT_FILE_ENTRY].filePath === treeKey;
```
WitoDelnat, line 75: We were quite inconsistent with the tree key as sometimes it ended as the absolute file path and sometimes as <root> (e.g. on reload). For easiest mental model I made the treeKey the same as the fileMap key, basing myself on these docs:

src/models/fileentry.ts:8:
> the path of the file relative to the root folder - used as key in the fileMap

src/models/appstate.ts:82
>  /** maps filePath to FileEntry
>  * - filePath is relative to selected rootFolder
>  * - fileMap[**ROOT_FILE_ENTRY**] is the FileEntry for the rootFolder and it's **filePath is absolute**
>  */

src/components/organisms/FileTreePane/FileTreePane.tsx
```diff
@@ -434,7 +423,7 @@ const FileTreePane = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [highlightNode]);
 
-  useEffect(() => {
```
WitoDelnat, line 437: This would not run properly as the tree was not yet present, the code would run while we were showing "Get started by selecting a folder containing manifests, kustomizations or Helm Charts." and as a result it did nothing.

src/components/organisms/FileTreePane/FileTreePane.tsx
```diff
@@ -57,9 +58,12 @@ const createNode = (
   rootFolderName: string
 ): TreeNode => {
   const resources = getResourcesForPath(fileEntry.filePath, resourceMap);
+  const isRoot = fileEntry.name === ROOT_FILE_ENTRY;
+  const key = isRoot ? ROOT_FILE_ENTRY : fileEntry.filePath;
+  const name = isRoot ? rootFolderName : fileEntry.name;
 
   const node: TreeNode = {
-    key: fileEntry.filePath,
```
WitoDelnat, line 62: This caused an inconsistency between the tree key and the file map key which gave subtle bugs. I made the keys between both consistent.

src/components/organisms/FileTreePane/TreeItem.tsx
devcatalin: You are correct that for `fileMap[ROOT_FILE_ENTRY]` the `filePath` is actually the absolute path of the project.
This root entry should've been a separate property in the AppState but due to the fact that it's heavily used in many places, a refactor would take a lot of time. Maybe we should update the comment on the `fileMap` to reflect this better.

I think it's okay to use the fileMap key as the treeKey, it should be unique

devcatalin (approved):
Nice, it works a lot better!

Issue #1602, opened 2022-04-14
crash when switching projects


olensmar:
I frequently get the below when switching between projects or creating new ones - although I can't reproduce it consistently

<img width="951" alt="image" src="https://user-images.githubusercontent.com/1917063/163337083-b8c756b7-ba1e-4c0f-aca1-7824e1f837df.png">

erdkse:
This was also a bug that we see in Nucleus. Great that it is solved!

PR #1604, opened 2022-04-14
fix: fix switch project

WitoDelnat:
This PR fixes https://github.com/kubeshop/monokle/issues/1602

## Fixes

- Fix project switch

## How to test it

- Have two projects, one with excluded files.
- Swap between them
- Validate it no longer goes black

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/FileTreePane/FileTreePane.tsx
+++ b/src/components/organisms/FileTreePane/FileTreePane.tsx
@@ -150,7 +150,7 @@ const FileTreePane = () => {
   const selectedPath = useAppSelector(state => state.main.selectedPath);
   const selectedResourceId = useAppSelector(state => state.main.selectedResourceId);
   const expandedFolders = useAppSelector(state => state.ui.leftMenu.expandedFolders);
-  const uiState = useAppSelector(state => state.ui);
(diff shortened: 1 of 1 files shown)


olensmar (approved):
thank you!

src/components/organisms/FileTreePane/TreeItem.tsx
```diff
@@ -74,9 +74,8 @@ const TreeItem: React.FC<TreeItemProps> = props => {
 
   const isRoot = treeKey === ROOT_FILE_ENTRY;
   const root = fileMap[ROOT_FILE_ENTRY];
-  const node = fileMap[treeKey];
```
WitoDelnat, line 77: For future reference, this bug occurred because `fileMap` does not include excluded files, yet we have an option to Show/Hide excluded files. `node` will be `undefined` in this case and it will crash.

WitoDelnat: This assumption is incorrect and a new fix has been pushed that fixes the actual issue - an outdated `tree`.

PR #1608, opened 2022-04-14
Julia/feat/walkthrough

jmelnych:
This PR...

## Changes

-

## Fixes

-

## How to test it
Walkthrough steps appears when a user has installed or updated the Monokle version. To test it locally - remove lastSeenReleaseNotesVersion in config file: ~/Library/Application Support/monokle/config.json (For MacOS)
-

## Screenshots
image

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/App.tsx
```diff
@@ -177,8 +183,10 @@ const App = () => {
       if (!semver.valid(lastSeenReleaseNotesVersion) || semver.lt(lastSeenReleaseNotesVersion, version)) {
         setAppVersion(version);
         setShowReleaseNotes(true);
+        dispatch(handleWalkThroughStep(1));
       }
     });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
WitoDelnat, line 189: Nitpick: I dislike disabling eslint, maybe `useEffectOnce` ([docs](https://github.com/streamich/react-use/blob/master/docs/useEffectOnce.md)) is more descriptive of what we want?

src/components/molecules/SectionRenderer/SectionHeader.tsx
```diff
@@ -111,21 +115,45 @@ function SectionHeader(props: SectionHeaderProps) {
...
+                  $nameSize={sectionBlueprint.customization?.nameSize}
+                  $nameWeight={sectionBlueprint.customization?.nameWeight}
+                  $nameVerticalPadding={sectionBlueprint.customization?.nameVerticalPadding}
+                  $nameHorizontalPadding={sectionBlueprint.customization?.nameHorizontalPadding}
+                  $level={level}
+                  onClick={toggleCollapse}
+                >
+                  {name}
+                </S.Name>
+              </Popover>
```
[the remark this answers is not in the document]
devcatalin: I agree with Wito that the above snippets looks cleaner and I would also say that instead of hard-coding the WalkThrough logic/components into the `SectionRenderer` component, we should use the `customization` object from the `SectionBlueprint` to provide a wrapper for the `S.Name`

src/App.tsx
devcatalin: In this specific case I think yes, the `useEffectOnce` could be used, but there are cases where we disable the `exhaustive-deps` so that the useEffect is triggered only on some of the dependencies.

With that in mind and not related to this PR, I'm wondering if there are alternative ways to solve the other places where we disable this lint rule.

WitoDelnat: It depends a bit case-by-case, sometimes there is even no way around it and Monokle is not your average web application so I expect it here from time to time.

src/App.tsx
```diff
@@ -177,8 +183,10 @@ const App = () => {
       if (!semver.valid(lastSeenReleaseNotesVersion) || semver.lt(lastSeenReleaseNotesVersion, version)) {
         setAppVersion(version);
         setShowReleaseNotes(true);
+        dispatch(handleWalkThroughStep(1));
```
andreiv1992, line 186: Can we use enum values instead of numbers?

src/components/molecules/WalkThrough/WalkThrough.tsx
```diff
@@ -0,0 +1,43 @@
...
+import {handleWalkThroughStep} from '@redux/reducers/ui';
+
+import {wkContent} from './wkContent';
+
+import * as S from './styled';
+
+type WalkThroughProps = {
+  walkThrough: {
+    content: string;
+    currentStep: number;
```
andreiv1992, line 11: Same thing with enums

src/App.tsx
devcatalin: I'm also okay with strings instead of enums if we use a union type like `currentStep: "someStep" | "otherStep" | "coolStep"`

src/components/molecules/SectionRenderer/SectionHeader.tsx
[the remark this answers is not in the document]
jmelnych: @WitoDelnat , totally agree, thank you for providing your suggestion

src/App.tsx
jmelnych: @andreiv1992, agree! updated

WitoDelnat (approved):
Cool work Yulia! I really think new users will enjoy what you made.

PR #1610, opened 2022-04-14
fix: Nucleus initialization

devcatalin:
This PR...

## Changes

-

## Fixes

- the `appStarted` call should happen before any other call related to Nucleus, as per the discussion we had with their support team 

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


erdkse (approved):
lgtm!

PR #1617, opened 2022-04-15
Add policy resource scanning

WitoDelnat:
This PR adds resource scanning. It only includes real-time feedback of the current resource, the overview panel is out-of-scope for this PR.

Under the hood it uses Open Policy Agent policies that are compiled to web assembly. 

## Changes

- It scans resources in the background while editing and shows errors in the editor.

## How to test it

- Enable `resourceScanning` feature flag.
- Open a resource that violates a policy. For instance, include a container that does not have resource limits.
- Validate that the editor shows you an error.

## Screenshots

Screenshot 2022-04-15 at 14 03 37

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/models/k8sresource.ts
```diff
@@ -1,12 +1,15 @@
...
+import {SarifRule} from './policy';
+
 export type RefNode = {scalar: Scalar; key: string; parentKeyPath: string};
 
 type ResourceValidationError = {
   property: string;
   message: string;
   errorPos?: RefPosition;
   description?: string;
+  rule?: SarifRule;
```
WitoDelnat, line 12: The idea is to make errors with different types here.

src/models/k8sresource.ts
```diff
@@ -52,6 +55,9 @@ interface K8sResource {
   };
   /** result of schema validation */
   validation?: ResourceValidation;
+
+  /** result of policy validation */
+  issues?: ResourceValidation;
```
WitoDelnat, line 60: I'd like to refactor validation to maybe be more elaborate, I kept it split off now as validateResource overwrites the whole object and I do not validatePolicy of all resources on editing file for performance.

src/components/molecules/Monaco/codeIntel.ts
```diff
@@ -291,7 +328,70 @@ export async function applyForResource(
...
+    const rule = issue.rule!;
+    const message = [
+      createMarkdownString(`__${issue.message}:__ ${rule.longDescription.text} ${rule.help.text}`),
+    ].filter(isDefined);
+
+    return createGlyphDecoration(issue.errorPos?.line ?? 1, GlyphDecorationTypes.PolicyIssue, message);
+  });
+
+  const inline: monaco.editor.IModelDeltaDecoration[] = [];
+  // const inline = issues
```
WitoDelnat, line 350: This is some work in progress. I want to read up about Monaco in the near future to know better the capabilities and differences. I'll remove this later but wanted it tracked in git.

src/components/molecules/Monaco/editorConstants.ts
```diff
@@ -1,15 +1,21 @@
 import {monaco} from 'react-monaco-editor';
 
+import Colors from '@styles/Colors';
+
+export const MODEL_OWNER = 'monokle';
```
WitoDelnat, line 5: TODO: see whether having your own model identifier clashes with the yaml key. If you use `yaml` it does not work properly. I expect because the `monaco-yaml` package clears markers that we set as Monokle.

src/redux/thunks/loadPolicies.ts
```diff
@@ -0,0 +1,220 @@
...
+        type: AlertEnum.Warning,
+      })
+    );
+    return [];
+  }
+});
+
+const DEFAULT_CONFIGURATION: PolicyConfig = {enabled: true};
+
+const DEFAULT_TRIVY_PLUGIN: BasicPolicy = {
```
WitoDelnat, line 47: Not sure yet whether we want rule metadata embedded in the Rego policies or within our plugin configuration. I decided that it's not relevant for now and we can refactor it later. This was the easiest so I went with it.

src/redux/thunks/loadPolicies.ts
```diff
@@ -0,0 +1,220 @@
...
+        text: 'The container should drop all default capabilities and add only those that are needed for its execution.',
+      },
+      helpUri: 'https://kubesec.io/basics/containers-securitycontext-capabilities-drop-index-all/',
+      help: {
+        text: "Add 'ALL' to containers[].securityContext.capabilities.drop.",
+      },
+      properties: {
+        severity: 'low',
+        entrypoint: 'appshield/kubernetes/KSV003/deny',
+        path: '$container.securityContext.capabilities.drop',
```
WitoDelnat, line 71: The idea is that this can be a full path or work with special keywords. `$container` is necessary because one rule checks containers for `Pod`, `Deployment`, `Job` which have containers located at different locations. I notice all rules so far are actually on container so this is the most important to have but we can extend this later.

src/redux/services/validation.ts
```diff
@@ -42,6 +44,79 @@ function getErrorPosition(valueNode: ParsedNode, lineCounter: LineCounter | unde
...
+      errorPos,
+      rule,
+    };
+  });
+
+  return errors;
+}
+
+function determineContainerErrorPos(resource: K8sResource, container: string): RefPosition {
+  const supported = resource.kind === 'Deployment' || resource.kind === 'StatefulSet';
```
WitoDelnat, line 87: TODO: extend with Job and Pod

src/models/k8sresource.ts
```diff
@@ -88,6 +94,8 @@ interface RefPosition {
   line: number;
   column: number;
   length: number;
+  endLine?: number;
+  endColumn?: number;
```
WitoDelnat, line 98: TODO: make this better

src/components/molecules/Monaco/codeIntel.ts
```diff
@@ -291,7 +328,70 @@ export async function applyForResource(
...
+  return {decorations: newDecorations, disposables: newDisposables};
+}
+
+function applyPolicyIntel(resource: K8sResource): {
+  decorations: monaco.editor.IModelDeltaDecoration[];
+  markers: monaco.editor.IMarkerData[];
+} {
+  const issues = resource.issues?.errors ?? [];
+
+  const glyphs = issues.map(issue => {
```
WitoDelnat, line 340: More a functional rather than technical question. I think we do not want to show the glyphs in the gutter. I checked how VSC does it and they only do underlining on errors. We need to investigate the UX of many errors though as including descriptions to it will give a popup over your whole screen. (Most language errors have a single error so it's different).

src/components/molecules/Monaco/editorConstants.ts
WitoDelnat: good enough. It seems to not interfere but will need to test more thoroughly later.

src/components/molecules/Monaco/codeIntel.ts
WitoDelnat: Keeping both for now since selecting parent node for missing error will obfuscate error of a more detailled node otherwise.

src/models/k8sresource.ts
WitoDelnat: Resolving this to be tackled later.

src/models/k8sresource.ts
WitoDelnat: Resolving this to be tackled later.

src/models/k8sresource.ts
WitoDelnat: Resolving this to be tackled later.

erdkse (approved):
The code looks great! But I was lost at some places since this is a big PR. Let's merge it and test for unexpected errors.

PR #1620, opened 2022-04-18
fix: remove ui test slack notification

andreiv1992:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/.github/workflows/monokle-ui-tests.yml
+++ b/.github/workflows/monokle-ui-tests.yml
@@ -91,13 +91,6 @@ jobs:
           name: playwright-output
           path: test-output/**
 
-      - name: Notify slack failure
-        env:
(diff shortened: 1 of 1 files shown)


erdkse:
Why do we remove slack alerts? @andreiv1992

erdkse:
> > Why do we remove slack alerts? @andreiv1992
> 
> Because it is spamming, at least now when not all the prs might be update with the latest code. If you are ok with them we can leave them on

I prefer @devcatalin to decide it. I agree it is spamming and we are able to see test results on the pipeline but anyway It is not bothering me.

devcatalin:
Let's keep it for now...

PR #1621, opened 2022-04-18
fix: ui updates colors and borders

jmelnych:
This PR...

## Changes
Updated according to Julio Comments:
please change colors to:
- top: #101314
- left: #191F21
- the border colors of drop-downs in the top, which are gray-4, have them to gray-5
- the left tab, when highlighted, looks like it have a 2px radio. we need to take that out
- first time, when file explorer is open, the tab on the left does not appear highlighted. We need to highlight it so the tab look is there from scratch

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/PageHeader/ClusterSelection.styled.tsx
+++ b/src/components/organisms/PageHeader/ClusterSelection.styled.tsx
@@ -100,7 +100,7 @@ export const ClusterOutlined = styled(RawClusterOutlined)`
 export const ClusterStatus = styled.div`
   display: flex;
   align-items: center;
-  border: 1px solid ${Colors.grey3};
+  border: 1px solid ${Colors.grey5};
   border-radius: 4px;
   padding: 0px 8px;
 `;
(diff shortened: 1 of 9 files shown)


devcatalin:
Nice! @jmelnych @julioramosest I feel like the background color of the left menu icons is changing slower than the pane appears. Is there an animation on the left menu icon background? Can we make the background color change of the icon to be instant?

julioramosest:
@jmelnych @devcatalin Thanks guys, colors look are correct. No strong opinion on color hover animation, I go along your comment Catalin. Just 1 thing pops up - since we now how the left arrow to collapse the column, maybe we don't need now the "Hide File explorer" etc tooltip. It appears just when you click to expand since the cursor is already there and it's a bit annoying. I would keep the action, just not the tooltip. We'll talk it over tomorrow in the UX meeting.

jmelnych:
@devcatalin, nice catch about left-panel button background - I removed the transition

julioramosest:
Right @jmelnych I think we can remove the "hide [xxx]" tooltips. And to have it perfect, I'd do one last thing. When highlighted, the tabs look a bit narrow vertically to me. I think if we could add 4px padding (or similar to that effect) above and below each icon they would look great. Also you can save that empty space there is now between label and label. What do you think?

julioramosest:
Now it's fine for me. Approving it.

devcatalin (approved):
Thank you, looks good to me now!

PR #1633, opened 2022-04-20
docs: add draft policy validation

WitoDelnat:
This PR adds documentation about policy validation.

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test


WitoDelnat:
Images still have to be added but the feature is not yet ready. We can already prepare the docs and add these in at the last moment.

WitoDelnat:
I haven't run this locally so no idea how the table will render. Please let me know if I should validate this.

devcatalin (approved):

Issue #898, opened 2021-12-15
Enable Form editors in Cluster Mode


olensmar:
We currently disable the Form editors in cluster mode although the source is enabled to make changes and apply them back - Form editors should be enable also..

PR #1648, opened 2022-04-21
feat: edit toggle in cluster mode

stan-gabriel:
This PR...

## Changes
Add new Settings option: Allow editing resources

## How to test it
- 1 Go to Setting -> disable "Allow editing resources" under "Cluster Mode"
- 2 Enable "Cluster Mode"
- 3 Should not be able to edit resources

## Screenshots
image

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/molecules/Monaco/Monaco.tsx
```diff
@@ -268,18 +269,18 @@ const Monaco = (props: {diffSelectedResource: () => void; applySelection: () =>
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [editor, selectedResourceId, firstCodeLoadedOnEditor]);
 
-  // read-only if we're in preview mode and another resource is selected - or if nothing is selected at all
+  // read-only if we're in preview mode and another resource is selected - or if nothing is selected at all - or allowEditInClusterMode is false
   const isReadOnlyMode = useMemo(() => {
+    if (!settings.allowEditInClusterMode && previewType === 'cluster') {
```
devcatalin, line 274:
```suggestion
    if (isInPreviewMode && previewType === 'cluster' && !settings.allowEditInClusterMode) {
```

devcatalin (changes requested):

src/components/molecules/Monaco/Monaco.tsx
stan-gabriel: fixed

olensmar:
nice! would be great if we could fix #898 together with this - to be consistent with this added option

stan-gabriel:
> nice! would be great if we could fix #898 together with this - to be consistent with this added option
I will take a look and fix #898 as well

devcatalin:
Merging this now as I'd prefer having #898 in a separate PR

Issue #1635, opened 2022-04-20
Panel resizing faulty when filter panel is open


olensmar:
When I open the filter panel at the top of the Navigator the resizing logic of the panels is fault; trying to drag the either divider is "sluggish" and only works to the right

WitoDelnat:
Panel resizing has difficulties on multiple fronts:

1. ResizableBox does not nest well in fluid setups.
2. ResizableBox also explicitly requires both width and height which is tedious in fluid environments where we have to resort to a lot of useMeasure and passing heights around to children instead of allowing Grid/Flex to do its thing.
3. ResizableBox uses explicit width instead of percentages, this implicates that we do a lot of calculations in useEffects ourselves to get these percentages. We need multiple effects to reach a correct steady state (with renders that are faulty but that rerender very quickly inbetween).

I want to check with the team if there is something tying us deeply to this library as alternatives work better for setups that are fluid like ours. Below I made a POC with react-reflex, where I first showcase the old behavior and at 0:44 the new library is showcased. I think this is the way forward, but would like to check with the team before spending more time to make sure we align on a solution.

https://user-images.githubusercontent.com/7761005/165111925-2d7ca010-884e-4583-9e61-f02aca515195.mov

PR #1668, opened 2022-04-27
Wito/feat/rework panels

WitoDelnat:
This PR closes https://github.com/kubeshop/monokle/issues/1635. It refactors the resizable panels by using react-reflex instead of react-resizable.

Remarks

- I was not able to remove react-resizable as it's still used by modals where this flex approach does not work.
- I **temporarily** removed bottom-actions and right panel to make certain we can test the new release in time.

## Changes

- Toggling left panel now adds/removes from edit pane instead navigator pane. This should make it easier to start editing resources.
- Resizing navigator pane has a max width. After max-width is reached it will propagate dragging and move it along the screen. This should reduce user actions needed since you only need to drag one Splitter.

## Fixes

- Resize works properly when the resource filter is open.

## How to test it

- Go to a project and open the resource filter.
- Resize one of the panels.
- Validate that it properly resizes.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/PaneManager/PaneManagerSplitView.tsx
```diff
@@ -25,139 +20,61 @@ const TemplateManagerPane = React.lazy(() => import('@organisms/TemplateManagerP
...
+                {leftMenuSelection === 'helm-pane' && <HelmPane />}
+                {leftMenuSelection === 'kustomize-pane' && <KustomizePane />}
+                {leftMenuSelection === 'templates-pane' && <TemplateManagerPane contentHeight={paneHeight} />}
+              </Suspense>
+            </S.LeftPane>
+          </ReflexElement>
         )}
 
-        <S.Pane id="NavPane" $height={paneHeight} ref={navPaneRef}>
+        {leftActive && leftMenuSelection && <ReflexSplitter propagate />}
```
WitoDelnat, line 61: 📓 just to spread some awareness: You can't use fragments within ReflexElement so you need to do this separately instead of grouping these within a fragment.

src/components/organisms/PaneManager/PaneManagerSplitView.tsx
```diff
@@ -25,139 +20,61 @@ const TemplateManagerPane = React.lazy(() => import('@organisms/TemplateManagerP
...
-
-        {featureJson.ShowGraphView && rightMenuSelection === 'graph' && rightActive && (
-          <S.Pane id="RightPane" $height={paneHeight}>
-            <GraphView editorHeight={paneHeight} />
-          </S.Pane>
-        )}
-      </S.SplitViewGrid>
+        </ReflexElement>
+
+        <ReflexSplitter propagate={leftActive && leftMenuSelection !== undefined} />
```
WitoDelnat, line 72: 📓 just to spread some awareness: you can only use propagate when you have multiple splitters or it gives unexpected behavior.

src/components/organisms/NavigatorPane/NavigatorPane.tsx
```diff
@@ -115,31 +103,27 @@ const NavPane: React.FC = () => {
...
-            height={height || 350}
-            axis="y"
-            resizeHandles={['s']}
-            minConstraints={[100, 200]}
-            maxConstraints={[width, paneHeight - 300]}
-            handle={(h: number, ref: LegacyRef<HTMLSpanElement>) => <span className="custom-handle" ref={ref} />}
-          >
+      <ReflexContainer orientation="horizontal" style={{height: height - 40}}>
+        {isResourceFiltersOpen && (
+          <ReflexElement flex={0.4} minSize={100}>
```
WitoDelnat, line 113: For now, I do not persist the resource filter's height.

src/components/organisms/PaneManager/PaneManagerSplitView.tsx
devcatalin: I would add this as a comment next to the code because it's useful to know

src/components/organisms/PaneManager/PaneManagerSplitView.tsx
devcatalin: I would add this as a comment as well

PR #1676, opened 2022-04-28
Wito/fix/unsaved resources

### LINKED ISSUE #1644: Bugs related to saving unsaved resources to a folder
There are a couple of things that do not work well and are related to the New Resource Wizard.

Steps to reproduce them:
1. Create a new project using the Basic Service Deployment template
2. Click on the checkboxes next to the service and deployment resource names in the Navigator
3. Use the "Save to file/folder" action

The first bug can be seen in the modal:
image

The generated file names are incorrect:
- the file name for the deployment should have been `test-deployment.yml`
- the file name for the service should have been `test-service.yml` (Note the lowercase `S`)

After clicking on Ok, we can see the files in the File Explorer:

image

Here the problem is that the files look like they are greyed out, but we can still click them.
We should find out why they look like that.

Furthermore, open the New Resource modal from the Navigator titlebar, enter a name and select a resource kind.
Then on the last option instead of "Don't save", select the "Add to file" option.

Now, the dropdown next to "Add to file" should list the existing yml files that we just created, but instead, it's empty:

(removed screenshot as it was not the correct one...)

If you can create a new resource in the navigator but use the "Don't save" option, then check the newly created resource and use the "Save to file/folder" action from the Navigator's you will see a similar behavior. (Choose the "append to file" option)

image

Finally, if you use the context menu on one of those greyed out files from the File Explorer and then click on the "Delete" action, the app will enter an error state and the screen will turn black.

WitoDelnat:
This PR fixes #1644.

## Fixes

- Adds resource kind to all resources on name clash.
- Fix an issue where newly added resources are unsupported.
- Fix an issue where newly added resources are unavailable in drop downs.
- Fix appending resources to a file.
- Fix reveal root.

## How to test it

- See the issue description.

## Screenshots

note: there is still a small issue when the appended file is highlighted but its not trivial to fix. These are already sufficient fixes and we can improve the file explorer later (Reason is that highlighted node is not in store).

https://user-images.githubusercontent.com/7761005/165745406-d25075b6-659d-4b1b-9b6f-ab6c62b43c40.mov

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/SaveResourcesToFileFolderModal/SaveResourcesToFileFolderModal.tsx
+++ b/src/components/molecules/SaveResourcesToFileFolderModal/SaveResourcesToFileFolderModal.tsx
@@ -54,11 +54,10 @@ const generateFullFileName = (
   existingFileNames: string[],
   includeKind?: boolean
 ): string => {
-  const {kind, name} = resource;
-  let fullFileName = getFullFileName(
-    `${name}${includeKind ? `-${kind}` : ''}${suffix ? ` (${suffix})` : ''}`,
-    fileIncludes
-  );
+  const name = resource.name;
+  const nameKind = includeKind ? `-${resource.kind.toLowerCase()}` : '';
+  const nameSuffix = suffix ? ` (${suffix})` : '';
+  const fullFileName = getFullFileName(`${name}${nameKind}${nameSuffix}`, fileIncludes);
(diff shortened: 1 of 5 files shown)


devcatalin (approved):
Great!

PR #1691, opened 2022-05-02
feat: separate walkthrough for new users and release features

jmelnych:
This PR...

Separates walkthrough logic for new users and new release features.
Currently, the new release feature contains only one step about the validation policy.
I removed the step about validation for a basic walkthrough guide, but maybe we want to join it all together?
Also, the replay functionality will replay only the basic guide.

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -192,7 +192,9 @@ const App = () => {
   const onCloseReleaseNotes = useCallback(() => {
     setShowReleaseNotes(false);
     if (!electronStore.get('appConfig.lastSeenReleaseNotesVersion')) {
-      dispatch(handleWalkThroughStep(StepEnum.Next));
+      dispatch(handleWalkThroughStep({step: StepEnum.Next, collection: 'novice'}));
+    } else {
+      dispatch(handleWalkThroughStep({step: StepEnum.Next, collection: 'release'}));
     }
     electronStore.set('appConfig.lastSeenReleaseNotesVersion', appVersion);
   }, [appVersion, dispatch]);
--- a/src/components/molecules/WalkThrough/types.ts
+++ b/src/components/molecules/WalkThrough/types.ts
@@ -1,3 +1,4 @@
+export type WalkThroughCollection = 'novice' | 'release';
 export type WalkThroughStep = 'template' | 'resource' | 'syntax' | 'cluster' | 'kustomizeHelm' | 'validation';
 export type WalkThroughContentProps = {
   data: {
@@ -5,6 +6,7 @@ export type WalkThroughContentProps = {
     title: string;
     content: string;
   };
+  collection: WalkThroughCollection;
   currentStep: number;
 };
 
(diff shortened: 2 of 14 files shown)


src/components/molecules/WalkThrough/types.ts
```diff
@@ -1,10 +1,12 @@
+export type WalkThroughCollection = 'novice' | 'release';
 export type WalkThroughStep = 'template' | 'resource' | 'syntax' | 'cluster' | 'kustomizeHelm' | 'validation';
```
WitoDelnat, line 2: A problem is that we can mix and match collections and steps in incorrect combinations. For example, you can set `<Walkthrough collection="release" step="template" />` which is not supported. Fixing it is a three step process.

1. Make step conditional based on the collection:
```
export type WalkThroughStep<C extends WalkThroughCollection = WalkThroughCollection> = C extends 'novice'
  ? 'template' | 'resource' | 'syntax' | 'cluster' | 'kustomizeHelm'
  : 'validation';
```
The default is there to avoid it complains everywhere but it gives you the option to get a narrowed result when you do add the collection.


2. Make a connection between the passed collection and steps:
```
// src/components/molecules/WalkThrough/WalkThrough.tsx:16
type WalkThroughProps<C extends WalkThroughCollection> = {
  step: WalkThroughStep<C>;
  children: React.ReactNode;
  placement?: PopoverProps['placement'];
  collection: C;
};
```

3. Allow the prop to be inferred automatically:
```
// src/components/molecules/WalkThrough/WalkThrough.tsx:71
const WalkThrough = <C extends WalkThroughCollection>(props: WalkThroughProps<C>) => {
```

4. Validate that it works. API consumption does not require declaring the generic as it can infer this.
```
//src/components/organisms/NavigatorPane/ClusterCompareButton.tsx:31~32
// error - Type '"validation"' is not assignable to type '"template" | "resource" | "syntax" | "cluster" | "kustomizeHelm"'.
 <WalkThrough placement="leftTop" step="validation" collection="novice">

```

jmelnych: @WitoDelnat, good point! I've updated types

src/App.tsx
```diff
@@ -192,7 +192,9 @@ const App = () => {
   const onCloseReleaseNotes = useCallback(() => {
     setShowReleaseNotes(false);
     if (!electronStore.get('appConfig.lastSeenReleaseNotesVersion')) {
-      dispatch(handleWalkThroughStep(StepEnum.Next));
+      dispatch(handleWalkThroughStep({step: StepEnum.Next, collection: 'novice'}));
+    } else {
+      dispatch(handleWalkThroughStep({step: StepEnum.Next, collection: 'release'}));
```
WitoDelnat, line 197: I wonder if this should be reset to `0` for each release? We'd be responsible for properly updating the content. Currently the only option would be to keep adding content to the collection, but then after your first update you get all the content of the previous releases which seems unintended.

jmelnych: Yes, we need to update release collection content along with release notes.
If we release only a patch version with no features, probably we don't want to show anything at all. We need to discuss it with the team

devcatalin (approved):

PR #1692, opened 2022-05-02
fix: fix left menu layout height

WitoDelnat:
This PR adds more layout improvements.

## Fixes

- Add scrolling when there are many Kustomize or Helm items. Previously it would just act as `overflow: hidden`.
- Fix several off by one pixels errors.
- Improve consistency across panels

## How to test it

- Open a project with many Helm charts or Kustomize directories.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


WitoDelnat:
I wonder if we should make the file explorer scroll box include the RootFolderText. For template, it makes sense to not include it as you want to be able to interact with the search box but this is not the case for that text. It makes the text disappear at a bit a weird place. What do you think?

Screenshot 2022-05-02 at 11 43 22

Screenshot 2022-05-02 at 11 39 28

devcatalin (approved):

PR #1729, opened 2022-05-05
feat: show keyboard shortcuts modal instead of link from help and men…

jmelnych:
…u panel

This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots
Screenshot 2022-05-05 at 07 38 47

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


jmelnych:
Grouped navigation vs tools shortcuts at my own discretion, wondering what you think guys

src/components/organisms/KeyboardShortcuts/KeyboardShortcuts.tsx
```diff
@@ -0,0 +1,151 @@
...
+  {name: 'Diff Resource', value: defineHotkey(hotkeys.DIFF_RESOURCE)},
+  {name: 'Reset Resource Filters', value: defineHotkey(hotkeys.RESET_RESOURCE_FILTERS)},
+];
+
+type BoardKeysProps = {
+  bindings: string;
+};
+
+const BoardKeys = ({bindings}: BoardKeysProps) => {
+  const commands: string[] = bindings.split(',').filter((item: string) => item);
```
topliceanurazvan, line 51: Maybe here we could have this in a separate file just to keep each component in its own separate file? I think it might be easier to follow the functionality 🤔

src/components/organisms/KeyboardShortcuts/styled.tsx
```diff
@@ -0,0 +1,80 @@
+import {Modal} from 'antd';
+
+import styled from 'styled-components';
+
+import Colors from '@styles/Colors';
+
+export const StyledModal = styled(Modal)`
+  .ant-modal-close-icon {
```
topliceanurazvan, line 8: Is this Modal different than the others used? Maybe we could modify the global antd theme in order to keep the modals consistent, wdyt?

src/components/organisms/KeyboardShortcuts/styled.tsx
```diff
@@ -0,0 +1,80 @@
...
+  width: 100%;
+  border-collapse: separate;
+  border-spacing: 0 10px;
+  overflow: hidden;
+  .ant-typography {
+    display: block;
+  }
+`;
+
+export const ShortCutContainer = styled.tr``;
```
topliceanurazvan, line 61: Are we going to add styles here? Otherwise I think we can remove it

src/components/organisms/KeyboardShortcuts/KeyboardShortcuts.tsx
jmelnych: good idea! moved it to its own component

src/components/organisms/KeyboardShortcuts/styled.tsx
jmelnych: yes, we can move it to a shared stylesheet

jmelnych:
@topliceanurazvan , thank you for your feedback, updated my PR

topliceanurazvan (dismissed):
Great work 👍🏻 - just one more thing I've noticed, the scrollbar styling seems to not be applied 

image

WitoDelnat:
This is super cool! I'm still reviewing the code but I already have a product request. It's a common shortcut that `shift + /` (so basically typing `?` character) opens this kind of menu. Could we add this here as well?

jmelnych:
> 
@WitoDelnat,
You mean, `shift + /` functionality to open a shortcuts itself?

src/components/organisms/KeyboardShortcuts/KeyboardShortcuts.tsx
```diff
@@ -0,0 +1,131 @@
...
+import Keyboard from '@assets/Keyboard.svg';
+
+import BoardKeys from './BoardKeys';
+
+import * as S from './styled';
+
+const {Text} = Typography;
+
+const shortcutsNavigation = [
+  {name: 'Navigate Back', value: defineHotkey(hotkeys.SELECT_FROM_HISTORY_BACK)},
```
WitoDelnat, line 19: I have one 'API' remark but I think it might be more work than is worth the effort. Hotkeys are something we'll regularly want to add so the design should make it easy to extend without having to go across multiple files.

```
export type Hotkey = keyof typeof HOTKEYS;

export const HOTKEYS = createHotkeys({
  SELECT_FROM_HISTORY_BACK: {
    name: 'test',
    key: 'cmd + h',
    category: 'tool',
  },
});

type HotkeyConf = {
  name: string;
  key: string;
  category: 'navigation' | 'tool';
};

function createHotkeys<TName extends string>(config: Record<TName, HotkeyConf>) {
  return config;
}
```

Then the other pages should be dynamic to avoid having to edit multiple files when adding a key:

```
const shortcutsNav = entries(HOTKEYS)
  .filter(([_, hotkey]) => {
    return hotkey.category === 'navigation';
  })
  .map(([id, hotkey]) => {
    return {name: hotkey.name, value: defineHotkey(hotkey.key)};
  });
```

WitoDelnat:
@jmelnych I tried playing around with both Gmail and Linear app and they use "?" and not "shift + /" (for my international keyboard layout that just happens to be "?" but if I change to Italian layout the hotkey changes. Not sure if we can do that kind of behaviour. Otherwise `shift + /` should already cover most Qwerty users. What do you think?

jmelnych:
@WitoDelnat, I like your idea, I think we can go with `shift + /`

jmelnych:
> Great work 👍🏻 - just one more thing I've noticed, the scrollbar styling seems to not be applied
> 
> image

Updated

WitoDelnat (approved):
Looks great

Issue #1735, opened 2022-05-05
Enabled OPA Validation for preview resources


olensmar:
Unfortunately OPA validation is currently disabled for resources that are generated by a preview - which is a shame as I would really like to be able to validate the output of my Helm charts and Kustomizations before deploying them.

Issue #1695, opened 2022-05-02
[validation] validate policies in Kustomize resources 


WitoDelnat:
Currently policy validation skips resources that are used within a Kustomization file. The difficulty is that these resources can be partial due to patches.

Let's illustrate with an example:

```
# demo-deployment
apiVersion: apps/v1
kind: Deployment
metadata:
  name: demo
spec:
  selector:
    matchLabels:
      app: demo
  template:
    metadata:
      name: demo
    spec:        
      containers:
        - name: server
           image: nginx-ingres:v2.0
```

And a patch:

```
# demo-patch-cpu
apiVersion: apps/v1
kind: Deployment
metadata:
  name: demo
spec:
  selector:
    matchLabels:
      app: demo
  template:
    metadata:
      name: demo
    spec:        
      containers:
        - name: server
           resources:
            limits:
              cpu: "150m"
```

If a Kustomization file puts these together then our editor will incorrectly say that the CPU limit is missing in the first example.

One idea is to show validation within a preview. The challenge will be to give actionable help. The preview is readonly and it's difficult to tell the file from which it originated. We can show the errors in the editor on the file that contains the resource that is not a patch. The idea is that once the validation errors are in SARIF format we can use the `relatedLocations` for this to allow Monaco to redirect to this page from within a preview.

PR #1736, opened 2022-05-05
fix: enabled OPA validation for preview resources 

olensmar:
and resources referred to by kustomizations - #1735 

This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/redux/services/validation.ts
+++ b/src/redux/services/validation.ts
@@ -45,9 +45,6 @@ function getErrorPosition(valueNode: ParsedNode, lineCounter: LineCounter | unde
 }
 
 export function validatePolicies(resource: K8sResource, policies: Policy[]): ResourceValidationError[] {
-  if (resource.filePath.startsWith('preview://')) {
-    return [];
-  }
   if (isManagedByKustomize(resource)) {
     return [];
   }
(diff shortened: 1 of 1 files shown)


src/redux/services/validation.ts
```diff
@@ -75,10 +72,6 @@ function isManagedByKustomize(resource: K8sResource): boolean {
     return true;
   }
 
-  if (resource.refs?.some(hasKustomizationReference)) {
```
olensmar, line 78: this check excludes vanilla resources that are used in a kustomization so I removed it - couldn't come up with a scenario where those should be removed!?

topliceanurazvan (approved):

src/redux/services/validation.ts
WitoDelnat: It gives false positives. I wrote an example in #1695.

olensmar: thanks for clarifying @WitoDelnat - I understand now although I think this scenario is much less common than just including valid resources in your kustomization setup (which seems to be the common case), so I suggest we remove this check for now. I'd rather have the UI show false positives in the uncommon scenario vs false negatives in the common one.

Going forward we we could improve the exclusion logic to exclude resources that are used by kustomizations that contain patches (would have to be recursive) - in which case I also think we should mark in the UI if/why a resource was excluded from validation so users don't incorrectly believe their resources are valid when they actually aren't.

thoughts @devcatalin ? @chlete ?

chlete: I also agree with @olensmar with following the most common scenario and improve as we move forward.

PR #1741, opened 2022-05-05
fix: reprocess resources on preview

WitoDelnat:
This PR reprocesses resources on generating a preview.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/redux/services/validation.ts
+++ b/src/redux/services/validation.ts
@@ -2,7 +2,7 @@ import Ajv, {ValidateFunction} from 'ajv';
 import {get} from 'lodash';
 import {Document, LineCounter, Node, ParsedNode, isCollection, isNode} from 'yaml';
 
-import {K8sResource, RefPosition, ResourceRef, ResourceValidationError} from '@models/k8sresource';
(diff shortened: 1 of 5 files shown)


olensmar (approved):
LGTM!

src/redux/services/validation.ts
```diff
@@ -177,7 +174,7 @@ function determineContainerIndex(
 ): YamlPath {
   for (let i = 0; i < properties.length; i += 1) {
     const property = properties[i];
-    const containers: {name: string}[] = get(resource.content, prefix.concat(property), []);
+    const containers = get<{name: string}[]>(resource.content, prefix.concat(property), []) ?? [];
```
WitoDelnat, line 177: 📓 TIL Lodash's get defaultValue only runs on `undefined` and not `null`.

In this case, some Helm charts incorrectly set `null` instead of `[]` to `initContainers` which causes the next line to crash because of accessing `findIndex` on `undefined`. It's their Helm charts not following spec but we should handle it gracefully.

devcatalin (approved):

PR #1744, opened 2022-05-05
feat: add feature flags

WitoDelnat:
This PR adds a small utility for feature flags.

## Changes

- It adds a `<FeatureFlag />` which dynamically renders based on typed flags.
- It adds a basic `useFeatureFlags` hook to abstract the feature flag implementation (i.e. the JSON file).

### DIFF
--- a/src/utils/features.tsx
+++ b/src/utils/features.tsx
@@ -1,0 +1,29 @@
+import React from 'react';
+
+import featureJson from '@src/feature-flags.json';
+
+const FEATURES = createFeatureFlags(featureJson, {
+  ShowGraphView: false,
+  ShowRightMenu: false,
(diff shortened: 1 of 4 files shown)


src/utils/features.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+import featureJson from '@src/feature-flags.json';
+
+const FEATURES = createFeatureFlags(featureJson, {
+  ShowGraphView: false,
+  ShowRightMenu: false,
+  ActionsPaneFooter: false,
+});
```
WitoDelnat, line 9: Downside: you have to define it here + once you want to override it you add it to JSON file. (so twice).
Upside: It's completely timed and easier to remove flags.

PR #1748, opened 2022-05-06
feat: add hide unsupported files setting

topliceanurazvan:
## Changes

- Add the `Hide Unsupported files` settings

## Screenshots

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


WitoDelnat:
Implementation is great!

From UX point of view I wonder if we should indicate somewhere that X files are hidden? You might've disabled unsupported files and forgotten and then wonder why some files are not showing up. I approved as this is something that can be added later onwards. It was just a random thought I had.

WitoDelnat (approved):
lgtm

topliceanurazvan:
That is a great suggestion @WitoDelnat , we could definitely show the number of hidden files somewhere. @julioramosest wdyt?

PR #1755, opened 2022-05-09
draft: Wito/fix/fix config merge

WitoDelnat:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/redux/services/projectConfig.ts
```diff
@@ -158,32 +158,8 @@ export const updateProjectSettings = (dispatch: (action: AnyAction) => void, pro
...
-  // add missing base properties
-  Object.keys(serializedConfig).forEach((key: string) => {
-    if (_.isUndefined(serializedBaseConfig[key])) {
-      serializedBaseConfig[key] = serializedConfig[key];
-    }
-  });
-
-  return flatten.unflatten(serializedBaseConfig);
+  invariant(baseConfig && baseConfig.settings && baseConfig.kubeConfig, 'base config expected');
+  return mergeWith(baseConfig, config, (_a, b) => (isArray(b) ? b : undefined));
```
WitoDelnat, line 162: This should fix crash due to incorrect merging due to parent in the middle not existing in base config.

```
// base.kubeConfig.context is undefined
// conf.kubeConfig.kubeConfig.contexts.0 // is defined
```

`flatten.unflatten` will fail while `mergeWith` will not.

electron/app/initKubeconfig.ts
```diff
@@ -61,6 +61,7 @@ export const getKubeConfigContext = (configPath: string): KubeConfig => {
     return {
       path: configPath,
       isPathValid: false,
+      contexts: [],
```
erdkse, line 64: This should actually fix all the problem - @WitoDelnat Could you please test?

erdkse:
#1848 #1843 #1811

electron/app/initKubeconfig.ts
WitoDelnat: Unfortunate I can't. I have no way to reproduce what I had. Simply ran into it once and noticed it was reported as well.

devcatalin:
I'd say we should merge this and see if the users still have the problem, looks like a good fix.

devcatalin (approved):

devcatalin:
I would keep the comments as well, maybe we should prefix them with `// TODO: ` ?

WitoDelnat:
I disabled compare everything and removed the TODOs. Should you need them you can look at commit 8858b05d36946ac314715fa36c4773f0486ecc76.

WitoDelnat:
It seems to happen when you did not have a cluster and afterwards start it. Likely the problem is that Monokle cannot handle changes to kube config properly. On latest release the following steps break, while on this branch they do not:

1. minikube stop
2. Start monokle
3. minikube start
4. Observe crash

Issue #1718, opened 2022-05-04
Improve the highlighting of Values properties in Templates (should handle more references))


chlete:
As a user I when I use the helm tool in Monokle, I'd like:

- [ ] When a Values reference is hovered, show the values in the popup
- [ ] The link to a values file should highlight and scroll to the line where that property is

olensmar:
nice! do we have an issue for the reverse functionality? i.e. something in the line of "when looking at a values file I would like to see a popup for each property that shows links to all templates that use the property" ?

chlete:
@olensmar your idea is here: #1754

PR #1756, opened 2022-05-09
Improve helm values

andreiv1992:
## Changes

- show value for helm values used in template files
- improve the regex for finding values in the helm file
- go to specific value in helm value file

## Fixes

- fix: #1718

## How to test it

-

## Screenshots

-
Screenshot 2022-05-09 at 15 28 44
Screenshot 2022-05-09 at 15 28 38

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/molecules/Monaco/codeIntel.ts
```diff
@@ -387,20 +389,65 @@ interface ApplyHelmFileArgs {
...
 }
 
 interface HelmValueMatch {
   path: string;
   keyPath: string;
+  value: any;
+  linePosition: RefPosition;
 }
 
+const get = (t: object, objPath: string) => objPath.split('.').reduce((r, k) => (r as any)?.[k], t);
```
devcatalin, line 402: any reason to implement this instead of using the `lodash.get` method?

src/components/molecules/Monaco/codeIntel.ts
```diff
@@ -387,20 +389,65 @@ interface ApplyHelmFileArgs {
...
+
+const getRange = (contents: any, keyPath: string): any => {
```
devcatalin, line 404: what "range" does this get? can we return a specific type instead of `any`?

devcatalin (changes requested):

PR #1778, opened 2022-05-12
refactor: remove global scrollbar style

topliceanurazvan:
## Changes

- Since the styling for the scrollbar was added in `index.css` there is no need for the `GlobalScrollbarStyle` anymore

## Fixes

- Overflow on x axis for Cluster Diff

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/utils/scrollbar.ts
```diff
@@ -1,15 +0,0 @@
-import Colors from '@styles/Colors';
-
-export const GlobalScrollbarStyle = `
```
WitoDelnat, line 3: I'm actually using this within the new compare modal now so I have to see how I can use your new approach. Could we pair for a bit before merging to see if it works?

WitoDelnat (approved):
LGTM!

PR #1779, opened 2022-05-12
fix: fix allow editing in cluster mode

stan-gabriel:
This PR...

## Fixes
Fix allow editing in cluster mode - True by default

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan (approved):

erdkse:
Do we want to save it to the `.monokle` file? I think that it is something we can do.

PR #1785, opened 2022-05-13
feat: images pane

topliceanurazvan:
## Changes

- Add Image Pane tab option
- Show list of found images in the current project
- Search functionality for images
- Highlight corresponding resources on image select
- Highlight corresponding images on resource select
- Add image outgoing link on refs popover
- Add image icon on Monaco editor image ref
- Select image on Monaco editor image ctrl+click
- Replace image tag functionality 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/PaneManager/PaneManagerLeftMenu.tsx
```diff
@@ -159,6 +162,27 @@ const PaneManagerLeftMenu: React.FC = () => {
         </WalkThrough>
       </PaneTooltip>
 
+      {DockerImagesPane && (
```
WitoDelnat, line 165: I'd recommend to instead use the following:

```
<FeatureFlag name="DockerImagesPane">
  ..
</FeatureFlag>
```

src/components/molecules/SectionRenderer/SectionHeader.tsx
```diff
@@ -35,7 +35,12 @@ function SectionHeader(props: SectionHeaderProps) {
     expandSection,
     collapseSection,
   } = props;
+
+  const {customization = {}} = sectionBlueprint;
```
devcatalin, line 39: could you revert the changes on this file? I remember it's not needed anymore and I'm currently refactoring the section blueprint components in a separate branch

src/components/molecules/SectionRenderer/SectionRenderer.tsx
```diff
@@ -208,11 +208,11 @@ function SectionRenderer(props: SectionRendererProps) {
         collapseSection={collapseSection}
         indentation={sectionIndentation || 0}
       />
-      {sectionInstance &&
-        sectionInstance.isVisible &&
-        !isCollapsed &&
-        itemBlueprint &&
-        sectionInstance.groups.length === 0 &&
+      {sectionInstance?.isVisible &&
```
devcatalin, line 211: same for this file

devcatalin (changes requested):

src/components/molecules/SectionRenderer/SectionRenderer.tsx
```diff
@@ -225,7 +225,13 @@ function SectionRenderer(props: SectionRendererProps) {
             options={itemRendererOptions}
             indentation={sectionIndentation || 0}
           />
-        ))}
+        ))
+      ) : (
+        <S.EmptyGroupText>
+          {sectionBlueprint.customization?.emptyVisibleItemsText || 'No items found.'}
```
WitoDelnat, line 231: Is it intended that this impacts the resource navigator?

Screenshot 2022-05-19 at 09 17 03

WitoDelnat:
I saw something passing in `dev` channel so I think you are aware of this already but this does not work with Kustomize.

Screenshot 2022-05-19 at 09 19 10

WitoDelnat:
Nitpick: Image gutter icon and error does not play together well

Screenshot 2022-05-19 at 09 21 10

WitoDelnat:
Some minor problems with linking

https://user-images.githubusercontent.com/7761005/169237882-4ccfc5f8-5e98-4901-a040-875693e20328.mov


https://user-images.githubusercontent.com/7761005/169237827-d9f24eab-34fc-4b76-b788-2f4e75e967d8.mov

devcatalin:
> I saw something passing in `dev` channel so I think you are aware of this already but this does not work with Kustomize.
> 
> Screenshot 2022-05-19 at 09 19 10

@WitoDelnat on this one we were not sure how to match the image since there is a `name` field but you can also provide `newName` and `newTag` fields, which would override the `name`.
So, in the Images panel, we're not sure if we should show only what's in the `name` field (easier solution) , or to actually do the replacement on the name. I'm leaning towards the first option, what do you think?

WitoDelnat:
> > I saw something passing in `dev` channel so I think you are aware of this already but this does not work with Kustomize.
> > Screenshot 2022-05-19 at 09 19 10
> 
> @WitoDelnat on this one we were not sure how to match the image since there is a `name` field but you can also provide `newName` and `newTag` fields, which would override the `name`. So, in the Images panel, we're not sure if we should show only what's in the `name` field (easier solution) , or to actually do the replacement on the name. I'm leaning towards the first option, what do you think?

@devcatalin 

For context, I imagine this the main use case for newName: it's mainly about changing registry. So above I should actually have used `image: keat-server` as an alias and then put the proper registry `newName: gcr.io/keatproj/keat-server` to easily modify this later.

For the list, I'd show `name` in the list. The `newName` is imo an implementation detail which you can view in the Kustomization file yourself. One nice touch would be an indicative label. For the example above: `keat-server (aliased)` where (aliased) is in a small grey font.

As for the action, I'd start with `Replace` to change both kustomization `name` and resource `image` to be in line with the vanilla replace functionality. If needed we can later add a `Replace name` specially for Kustomize as I don't think the `newName` is used that often anyway.

Does that make sense to you?

WitoDelnat (approved):

devcatalin (approved):

Issue #1754, opened 2022-05-09
Helm: Display relationship from Values properties to Templates


chlete:
### Description
As a user I would like when looking at a values, to see a popup for each property that shows links to all templates that use the property.

### Criteria

- [ ] While mouse over the values properties a tooltip will display the relationships to all templates used by this property

PR #1790, opened 2022-05-16
helm display relationship from values properties to templates

andreiv1992:
This PR...

## Changes

-

## Fixes

- fix: #1754 

## How to test it

-

## Screenshots
Screenshot 2022-05-16 at 08 23 47
Screenshot 2022-05-16 at 08 23 27
Screenshot 2022-05-16 at 08 23 08
Screenshot 2022-05-16 at 08 22 56

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/redux/services/helm.ts
```diff
@@ -79,20 +82,81 @@ export function isHelmChartFolder(files: string[]): boolean {
...
+  const filePath = getAbsoluteFilePath(fileEntry.filePath, fileMap);
+  const fileContent = fs.readFileSync(filePath, 'utf8');
+  const lineCounter = new LineCounter();
+  const documents = parseAllYamlDocuments(fileContent, lineCounter);
+
+  const values: HelmValueMatch[] = [];
+  documents.forEach((doc: any) => {
+    const helmObject = doc.toJS();
+    getObjectKeys(helmObject).forEach(keyPath => {
+      const scalar = getYamlScalar(doc.contents, keyPath);
```
WitoDelnat, line 121: I think you use the build-in `doc.getIn(KeyPath, true)` instead.

src/redux/services/helm.ts
```diff
@@ -208,3 +278,55 @@ export function findContainingHelmCharts(helmChartMap: HelmChartMapType, fileEnt
     });
   return charts;
 }
+
+export const getObjectKeys = (obj: any, prefix = ''): string[] =>
```
WitoDelnat, line 282: We have a library called `flat` installed. I think it does the same as you are trying to do here yourself.

See https://www.npmjs.com/package/flat.

WitoDelnat (approved):
Nice work, just a few minor nitpicks.

src/redux/services/helm.ts
andreiv1992: Thanks for the library, it would work usually but if I have an object like this:

`{
    key1: {
        keyA: 'valueI'
    },
    key2: {
        keyB: 'valueII'
    },
    key3: { a: { b: { c: 2 } } }
}`
I expect it to return 
`['key1', 'key1.keyA']`

Basically I want it to return all the elements not just the bottom leafs. I need this because in some cases we have object or array references in helm templates

WitoDelnat: Would you mind adding a small JavaDoc with this example? That way its intend is clear and what makes it special.

src/redux/services/helm.ts
andreiv1992: Tried that but it does not work as expected :(

PR #1795, opened 2022-05-17
Wito/feat/analytics opa

WitoDelnat:
This PR tracks OPA validation events.

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/utils/telemetry.ts
+++ b/src/utils/telemetry.ts
@@ -4,11 +4,14 @@ import Nucleus from 'nucleus-nodejs';
 
 import {isRendererThread} from './thread';
 
-export const trackEvent = (eventName: string, payload?: any) => {
+export const trackEvent = <TEvent extends Event>(
+  eventName: TEvent,
+  ...payload: EventMap[TEvent] extends undefined ? [undefined?] : [EventMap[TEvent]]
+) => {
   if (isRendererThread()) {
     ipcRenderer.send('track-event', {eventName, payload});
   } else {
-    Nucleus.track(eventName, payload);
+    Nucleus.track(eventName, payload as any);
   }
 };
 export const trackError = (error: any) => {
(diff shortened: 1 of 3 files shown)


src/utils/telemetry.ts
```diff
@@ -19,6 +24,35 @@ export const trackError = (error: any) => {
...
+  CLUSTER_COMPARE: {numberOfResourcesBeingCompared: number} | {fail: string};
+  FOLLOW_LINK: {type: string};
+  QUICK_SEARCH: undefined;
+  UPDATE_APPLICATION: undefined;
+  APPLY_FILE: undefined;
+  APPLY_HELM_CHART: undefined;
+  RUN_PREVIEW_CONFIGURATION: undefined;
+  VALIDATION_PANE_OPENED: {id: ValidationIntegrationId};
+  OPA_ENABLED: {all: boolean};
+  OPA_DISABLED: {all: boolean};
```
WitoDelnat, line 53: I wonder if our analytics platform works well enough to work with these kind of parametrised events or whether we need to split it up in OPA_ENABLED and OPA_ALL_ENABLED.

devcatalin: it works well, we can send any object as a payload as long as it doesn't contain functions

src/utils/telemetry.ts
```diff
@@ -2,13 +2,18 @@ import {ipcRenderer} from 'electron';
 
 import Nucleus from 'nucleus-nodejs';
 
+import {ValidationIntegrationId} from '@components/organisms/ValidationDrawer/integrations';
```
devcatalin, line 5: here we might have the problem with importing `.tsx` to main process?

WitoDelnat: Hmm you are correct. Damnit, I just fixed that in my other branch. Really annoying limitation. What do you think if I add jsx to TS compiler and we only use `import type`? I think it's unlikely anyone will use React components in Node process.

WitoDelnat: Actually this is just a `.ts` file so it works properly. We can reconsider this if we actually run into it again but for this PR it is fine.

devcatalin (approved):
LGTM

PR #1797, opened 2022-05-17
Wito/feat/add resource sets

WitoDelnat:
This PR adds all resource set types.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/CompareModal/CompareModal.tsx
```diff
@@ -34,17 +35,24 @@ export const DiffModal: React.FC<Props> = ({visible, onClose}) => {
...
+        </Col>
+      </Row>
 
+      <div
+        style={{
+          height: `calc(100% - ${height}px - 66px)`,
+          position: 'relative',
+          overflowY: 'auto',
+          overflowX: 'hidden',
+        }}
```
WitoDelnat, line 54: Please ignore all these `style` and layout things while reviewing. I'll come back and fix this properly later I add more functionality such as different operations and filter.

src/redux/reducers/compare.ts
```diff
@@ -359,27 +407,52 @@ export const selectComparisonListItems = createSelector(
...
+      predicate: (action, state) => {
         const resourceSetUpdated = isAnyOf(resourceSetSelected, resourceSetRefreshed, resourceSetCleared);
         if (!resourceSetUpdated(action)) return false;
+
         const actionSide = action.payload.side;
-        return actionSide === side;
+        if (actionSide !== side) return false;
+
+        const resourceSet = selectResourceSet(state.compare, side);
+        if (!isCompleteResourceSet(resourceSet)) return false;
```
WitoDelnat, line 443: So this is the whole essence of allowing this kind of user experience.

The configuration in our store is kept partially and only once it's completely present do we move on to the next stage.

src/redux/services/compare/fetchResources.ts
```diff
@@ -56,7 +70,78 @@ async function fetchResourcesFromCluster(state: RootState): Promise<K8sResource[
     const allYaml = fulfilledResults.map(r => (r as any).value).join(YAML_DOCUMENT_DELIMITER_NEW_LINE);
     return extractK8sResources(allYaml, CLUSTER_DIFF_PREFIX + String(kc.currentContext));
   } catch (err) {
-    log.debug('fetch resources form cluster failed', err);
+    log.debug('fetch resources from cluster failed', err);
+    throw err;
+  }
+}
+
+async function previewHelmResources(state: RootState, options: HelmResourceSet): Promise<K8sResource[]> {
```
WitoDelnat, line 78: I wonder how we can re-use this better.. I think the `thunk` and this `listener` should both use one underlying abstraction to fetch resources though it felt a bit out of scope for this. Wdyt?

devcatalin, line 78: we'll probably have to see if we want/can reuse some of this code

devcatalin (approved):

src/components/organisms/CompareModal/ResourceSetSelector.tsx
```diff
@@ -75,3 +68,154 @@ export const ResourceSetSelector: React.FC<Props> = ({side}: Props) => {
     </S.ResourceSetSelectorDiv>
   );
 };
+
+function ResourceSetTypeSelect({side}: Props) {
```
WitoDelnat, line 72: Shall I extract all these selectors to separate files? At that point I feel like the ResourceSetSelector should become its own directory.

devcatalin: yeah, I think that makes sense

PR #1801, opened 2022-05-18
Wito/feat/add operations and search

WitoDelnat:
This PR adds all operations, search and the possibility for an initial view.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/CompareModal/ComparisonList.tsx
+++ b/src/components/organisms/CompareModal/ComparisonList.tsx
@@ -1,4 +1,4 @@
-import {useCallback} from 'react';
+import {useCallback, useMemo} from 'react';
 
 import {Button, Checkbox, Col} from 'antd';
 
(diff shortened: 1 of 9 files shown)


src/components/organisms/CompareModal/ComparisonList.tsx
```diff
@@ -15,10 +15,25 @@ import {ComparisonItemProps, HeaderItemProps} from './types';
...
+        if (item.type === 'comparison') return true;
+        const nextItem = partiallySearchedItems[index + 1];
+        return nextItem && nextItem.type === 'comparison';
+      });
+  }, [items, search]);
 
   return (
     <div>
-      {items.map(item => {
+      {searchedItems.map(item => {
```
WitoDelnat, line 36: Filtering on search so late in the process makes this performant. Compared to the upcoming filters you no longer have to do any recalculations. The downside of this approach is that the headers keep showing the total resource count instead the filtered resource count. I think this is a decent trade-off but I wanted to highlight this as it could be considered a bug. What do you think?

src/components/organisms/NavigatorPane/CompareButton.tsx
```diff
@@ -7,13 +7,22 @@ import {SwapOutlined} from '@ant-design/icons';
...
+    const defaultView: ComparisonView = {
+      leftSet: {
+        type: 'local',
+      },
+      rightSet: {
+        type: 'cluster',
+        context: 'minikube',
+      },
+    };
+    dispatch(compareToggled({value: true, initialView: defaultView}));
```
WitoDelnat, line 25: The `initialView` seemed like a nice small feature that opens a lot of possibilities. Imagine you are previewing a Helm chart, on the other _values.yaml_ files we could on right-click add a compare that opens this view pre-filled.

This defaultView here will be removed later but is now for faster development feedback cycles.

src/redux/services/compare/filterResource.ts
```diff
@@ -0,0 +1,24 @@
+import {ResourceComparison} from '@redux/reducers/compare';
+
+type FilterOptions = {
+  search?: string;
+};
+
+type ResourceFilter = (value: ResourceComparison) => boolean;
```
devcatalin, line 7: would it make sense to rename this to `ResourceSetFilter` ? just to not mistakenly use the other `ResourceFilter` type that we have in the project

devcatalin (changes requested):

src/redux/services/compare/filterResource.ts
WitoDelnat: Oh yeah, I mistakenly used K8sResource in the beginning and forgot to rename this boy. I'll update it to `ComparisonFilter`.

devcatalin (approved):
LGTM

PR #1831, opened 2022-05-23
Wito/feat/add transfer

WitoDelnat:
This PR adds resource transfer (deploy to cluster / extract to local) for the selected resources.

Out of scope and coming in next PR: Transfer for single resources.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/CompareModal/CompareModal.tsx
```diff
@@ -36,25 +37,36 @@ export const DiffModal: React.FC<Props> = ({visible, onClose}) => {
...
-        <Col span={2} />
-        <Col span={11}>
+        <Col span={4} />
+        <Col span={10}>
           <ResourceSetSelector side="right" />
         </Col>
       </Row>
 
       <div
         style={{
```
topliceanurazvan, line 50: Can we have a styled component for this <div> ? Also why it is needed to use `calc` for the height?

src/components/organisms/CompareModal/CompareModal.tsx
```diff
@@ -36,25 +37,36 @@ export const DiffModal: React.FC<Props> = ({visible, onClose}) => {
...
+          marginRight: '-8px',
           position: 'relative',
           overflowY: 'auto',
           overflowX: 'hidden',
         }}
       >
         {status === 'selecting' ? <CompareModalSelecting /> : <CompareModalComparing />}
       </div>
+
+      <Row style={{height: 66, alignItems: 'center'}}>
```
topliceanurazvan, line 61: Also here we could have a styled Row

topliceanurazvan (commented):
Looks good to me 👍🏻 left a few comments

src/components/organisms/CompareModal/CompareModal.tsx
WitoDelnat: I still want to focus a bit on functionality before getting code style in order as things are still in flux a lot. I promise I'll extract this before its done.

As for the `calc`, It's needed because the resource set selector can increase in height but I need fixed height on this column to get proper overflow behaviour.

PR #1838, opened 2022-05-24
Compare everything: namespace override and improved inspection

WitoDelnat:
This PR adds default namespace override and improved inspection.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/CompareModal/CompareActionBar.tsx
+++ b/src/components/organisms/CompareModal/CompareActionBar.tsx
@@ -1,16 +1,21 @@
-import {useCallback} from 'react';
+import {useCallback, useMemo, useState} from 'react';
+import {useDebounce} from 'react-use';
 
-import {Checkbox, Input, Select, Space} from 'antd';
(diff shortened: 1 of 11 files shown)


devcatalin (approved):
LGTM

src/components/organisms/CompareModal/CompareActionBar.tsx
```diff
@@ -70,6 +63,61 @@ export const CompareActionBar: React.FC = () => {
   );
 };
 
+function SearchInput({disabled}: {disabled: boolean}) {
```
WitoDelnat, line 66: Probably I'll get a comment to split this up, but the total file length is now 170 lines which is a decent length  imo. The compareModal directory is getting rather big and I feel like splitting this up reduces readability instead of improving it. Rather personal (and maybe controversial) opinion so let me know what you think!

devcatalin: I saw it but looks fine, it's nice that you have it under the main component.

PR #1852, opened 2022-05-27
Wito/feat/compare wrapup

WitoDelnat:
This PR wraps up compare everything

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/redux/compare/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './listeners';
```
WitoDelnat, line 1: The huge compare.ts file had to be split up. I decided to put it under `redux/compare`. I'm not certain if this is the best approach since it's a first more isolated feature-slice.

src/redux/compare/state.ts
```diff
@@ -0,0 +1,143 @@
...
+      rightSet: undefined,
+    },
+    selection: [],
+    transfering: {
+      pending: false,
+    },
+  },
+};
+
+export type CompareState = {
```
devcatalin, line 19: I would add comments to the fields to explain how/why they are used, similar to how we do in `models/k8sresource.ts`

src/redux/compare/thunks.ts
```diff
@@ -0,0 +1,104 @@
...
+export const transferResource = createAsyncThunk<
+  {side: CompareSide; delta: MatchingResourceComparison[]; from: string; to: string},
+  TransferResourceArgs,
+  ThunkApi
+>('compare/transfer', async ({ids, direction}, {getState, dispatch}) => {
+  try {
+    const delta: MatchingResourceComparison[] = [];
+    const failures: string[] = [];
+
+    const namespace = getState().compare.current.view.namespace;
```
devcatalin, line 32: maybe you could do `const compareState = getState().compare;` ?

devcatalin (approved):

src/redux/compare/state.ts
```diff
@@ -0,0 +1,143 @@
...
+    view: ComparisonView | SavedComparisonView;
+    left?: ResourceSetData;
+    right?: ResourceSetData;
+    comparison?: ComparisonData;
+    selection: string[];
+    inspect?: ComparisonInspection;
+    search?: string;
+    filtering?: {
+      comparisons: ResourceComparison[];
+    };
```
WitoDelnat, line 32: I've been looking at https://redux-toolkit.js.org/api/createEntityAdapter and maybe we should consider adopting this pattern more.

Something along the lines of:
```
current: {
  comparisons: {
    ids: string[];
    filtered: string[];
    selected: string[];
    entities: ResourceComparison[];
  }
}
```

Issue #1880, opened 2022-05-31
Hovering on edit button in Cluster Selector doesn't highlight the row


topliceanurazvan:
When in Cluster Selector dropdown, if you try to hover the row ( context ) by hovering where the Edit button should be, the row doesn't get hovered.

## Screenshots

https://user-images.githubusercontent.com/47887589/171163633-f306a830-dc95-4c4a-ac1f-e9bda83e71e1.mp4

PR #1883, opened 2022-05-31
fix: hover on row - edit btn

stan-gabriel:
This PR...

## Fixes
https://github.com/kubeshop/monokle/issues/1880

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/PageHeader/ClusterSelectionTable.tsx
+++ b/src/components/organisms/PageHeader/ClusterSelectionTable.tsx
@@ -229,8 +229,9 @@ export const ClusterSelectionTable: FC<CLusterSelectionTableProps> = ({setIsClus
         <Column
           className="table-column-actions"
           key="clusterActions"
+          dataIndex="clusterActions"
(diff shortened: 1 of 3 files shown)


src/components/organisms/PageHeader/ClusterSelectionTable.tsx
```diff
@@ -229,8 +229,9 @@ export const ClusterSelectionTable: FC<CLusterSelectionTableProps> = ({setIsClus
         <Column
           className="table-column-actions"
           key="clusterActions"
+          dataIndex="clusterActions"
           ellipsis
-          width={70}
+          width={140}
```
topliceanurazvan, line 234: Why did we need to make the column bigger? Maybe we should leave it as 70 🤔

topliceanurazvan (changes requested):

topliceanurazvan (approved):

PR #1899, opened 2022-06-02
fix: deselect image on preview/exit

topliceanurazvan:
## Fixes

- Deselect image on preview / exiting preview

## How to test it

- Have an image selected and load the cluster

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


devcatalin:
You should remove the selected image when the project is changed as well. Probably needs to be handled in `setRootFolder.fulfilled`.
Alternatively, search for `selectedResourceId = undefined` to check if there are any other places where the image should be set to `undefined`.

devcatalin (approved):
LGTM

PR #1946, opened 2022-06-16
feat: validation drawer as tab

topliceanurazvan:
## Changes

- Make validation drawer as a left menu tab
- Scrollable table for the OPA rules
- Refactor validation redux store

## Screenshots

image

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/PaneManager/PaneManagerLeftMenu.tsx
```diff
@@ -94,7 +86,7 @@ const PaneManagerLeftMenu: React.FC = () => {
       >
         <MenuButton
           id="file-explorer"
-          isSelected={!leftDrawerVisible && leftMenuSelection === 'file-explorer'}
+          isSelected={Boolean(activeProject) && leftMenuSelection === 'file-explorer'}
```
andreiv1992, line 89: `leftMenuSelection === 'file-explorer'` => `leftMenuSelection === Enum.FileExplorer` something like this
I know that this was not your doing.. I prefer to use enums instead of strings even if they are mapped to types.
Enums offer a lot more context at a glance and it is a lot easier to find usages and improve code readability

src/components/organisms/PaneManager/PaneManagerLeftMenu.tsx
```diff
@@ -196,7 +188,7 @@ const PaneManagerLeftMenu: React.FC = () => {
         placement="right"
       >
         <MenuButton
-          isSelected={Boolean(activeProject) && !leftDrawerVisible && leftMenuSelection === 'templates-pane'}
+          isSelected={Boolean(activeProject) && leftMenuSelection === 'templates-pane'}
```
andreiv1992, line 191: For the sake of simplicity and core readability can we create a function which receives leftMenuSelection value as a param and we check the active project and === between current value and selected value. It seems that all of the are the same more or less

src/components/organisms/ValidationPane/OpenPolicyAgent/useOpenPolicyAgentTable.tsx
```diff
@@ -63,21 +63,21 @@ export function useOpenPolicyAgentTable() {
       },
       {
         key: 'severity',
-        title: 'Severity',
+        title: `${width < 420 ? '' : 'Severity'}`,
```
andreiv1992, line 66: Can we set 420 in constant somewhere so it's easier to change and can we have a descriptive name for it?

src/models/ui.ts
```diff
@@ -51,7 +49,13 @@ export type MonacoUiState = {
   selection?: MonacoUiSelection;
 };
 
-export type LeftMenuSelectionType = 'file-explorer' | 'helm-pane' | 'kustomize-pane' | 'templates-pane' | 'images-pane';
+export type LeftMenuSelectionType =
```
andreiv1992, line 52: This could be transformed in an enum

src/components/organisms/PaneManager/PaneManagerLeftMenu.tsx
topliceanurazvan: I will leave them unchanged for now, we can maybe do a tech debt refactoring of using Enums.

erdkse (approved):

devcatalin (approved):

Issue #1263, opened 2022-02-01
Show the file name to be created when adding a new resource and save to folder


topliceanurazvan:
We should show the file that is going to be created when saving to a folder a new resource and apply the same logic as for saving an unsaved resource.

![image](https://user-images.githubusercontent.com/47887589/152006042-374d20e9-23be-499e-9e5d-3cc3121e2024.png)

jenil777007:
Hello contributors, a first time contributor here, wanted to pick up this enhancement, do we have a reference of what the final result should look like? We should be using what component?

chlete:
Hi @jenil777007 thanks for contributing to Monokle! at this time we are closing a release, we are a bit busy, but we will get back to you!

jenil777007:
Absolutely @topliceanurazvan , been waiting to contribute. 

I have already asked my doubts above.

topliceanurazvan:
Basically, the name shown here should be similar to the one from saving an unsaved resource to a folder.

![image](https://user-images.githubusercontent.com/47887589/167147672-320cdaef-7de1-4a6d-8eec-fab1c538b4ad.png)

All the functionality it's in `SaveResourcesToFileFolderModal` component, especially `generateFullFileName` function which is taking care of generating the name for the resource you are trying to create. The naming convention is that, if the name already exists in the file system, it will just add the kind to the name ( e.g. `test-resource-secret` in our case ), and if this still exists, it will add some suffixes ( e.g. `test-resource-secret (2)`,`test-resource-secret (3)`, etc... ) until the generated name is unique.

About the label, having something similar like ( `File to be created:` and on the next row the actual name of the resource ), just like in the screenshot example.

Hopefully I was clear enough about what should be done and please reach out to us if you need more help and directions.

topliceanurazvan:
Hello @jenil777007 , how is it going with this issue? Is there anything else we could help with?

topliceanurazvan:
Great work @jenil777007 , don’t hesitate to ask us if you need any help.

topliceanurazvan:
> Hello @topliceanurazvan , I'm working on the filename generation logic, should have an update in a day or two.
> 
> As of now, I'm able to render the filename as followed,
> 
> ![Screenshot 2022-06-15 at 11 15 30 PM](https://user-images.githubusercontent.com/8577729/173891813-e1f4d27e-1d73-4a1d-a276-ae62091c15a6.png) .

[...]

PR #1948, opened 2022-06-16
Show the file name to be created when adding a new resource and save to folder #1263

jenil777007:
This PR...

## Changes
https://github.com/kubeshop/monokle/issues/1263
-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -27,12 +31,80 @@ import {openNamespaceTopic, openUniqueObjectNameTopic} from '@utils/shell';
...
+  includeKind?: boolean
+): string => {
+  const name = resourceName;
+  const nameKind = includeKind ? `-${resourceKind.toLowerCase()}` : '';
+  const nameSuffix = suffix ? ` (${suffix})` : '';
+  const fullFileName = fetchFullFileName(`${name}${nameKind}${nameSuffix}`);
+  let foundFile: fs.Dirent | FileEntry | undefined;
+
+  if (subfiles.length) {
+    foundFile = subfiles.find(dirent => dirent.name === fullFileName);
```
topliceanurazvan, line 60: We don't need to use the subfiles in here as we are saving just in our local folder, not outside the project, hence not needed to search for subfiles.

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -27,12 +31,80 @@ import {openNamespaceTopic, openUniqueObjectNameTopic} from '@utils/shell';
...
+  resourceName: K8sResource['name'],
+  resourceKind: K8sResource['kind'],
+  selectedFolder: string,
+  fileMap: FileMapType
+) => {
+  let subfiles: fs.Dirent[] = [];
+  const fsReaddirPromise = util.promisify(fs.readdir);
+  subfiles = (await fsReaddirPromise(selectedFolder, {withFileTypes: true})).filter(dirent => !dirent.isDirectory());
+
+  const hasNameClash = subfiles.map(x => x.name.split('.')[0]).some(subfile => subfile === resourceName);
```
topliceanurazvan, line 103: For this we should look through all the resources and find if there is another resource with the same name, similar to how it's done in `generateCreateaAndReplacedFiles` function from `SaveResourcesToFileFolderModal` component

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -230,6 +303,16 @@ const NewResourceWizard = () => {
         form.setFieldsValue({selectedResourceId: SELECT_OPTION_NONE});
       }
     }
+
+    if (savingDestination === 'saveToFolder') {
+      const generatedFileName = await generateFullFileName(
```
topliceanurazvan, line 308: This should be done in an `useEffect` , everytime the `selectedFolder` variable is changed

topliceanurazvan (changes requested):
Left a few comments. We don't need to use `fsReaddirPromise` as we already have all the paths in fileMap.

chlete:
hi @jenil777007 ! Thanks a lot for your contribution! Happy to know that you are using Monokle. Is there anything else that for you is missing that Monokle should have?

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
jenil777007: I have moved the logic to useEffect now

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
jenil777007: This has been amended

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
jenil777007: This has been amended now. However, I am still wondering if we're only checking name clashes in resource list, the name update will happen at project level, since resources constructed are available in resourceMap. Are we not checking for conflicts inside the selectedFolder?

topliceanurazvan: Completely agree, instead of searching name clashes through all the resource names, we should filter from the `resourceMap` only the resources whose `filePath` starts with the `selectedFolder`.

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -91,6 +129,32 @@ const NewResourceWizard = () => {
...
+
+      if (currentFolder) {
+        setSelectedFolder(currentFolder.slice(1));
+      } else {
+        setSelectedFolder(ROOT_FILE_ENTRY);
+      }
+      return;
+    }
+
+    const resourceNames = Object.keys(resourceMap).map(id => resourceMap[id].name);
```
topliceanurazvan, line 144: Could use `Object.values` directly here, but let's also filter them by `filePath` that starts with the `selectedFolder`

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -146,6 +210,14 @@ const NewResourceWizard = () => {
     setSubmitDisabled(!defaultValues?.name && !defaultValues?.kind && !defaultValues?.apiVersion);
   }, [defaultInput, defaultValues, isFolderOpen]);
 
+  useEffect(() => {
+    if (savingDestination !== 'saveToFolder') {
+      return;
+    }
+
+    generateExportFileName();
+  }, [selectedFolder]);
```
topliceanurazvan, line 219: Here we should also add the `savingDestination` as a dependency, otherwise the first time you switch to `saveToFolder` the name won't be generated

topliceanurazvan (changes requested):

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -91,6 +129,41 @@ const NewResourceWizard = () => {
...
+      } else {
+        setSelectedFolder(ROOT_FILE_ENTRY);
+      }
+      return;
+    }
+
+    let selectedFolderResources;
+    if (selectedFolder === ROOT_FILE_ENTRY) {
+      selectedFolderResources = Object.values(resourceMap).filter(
+        resource => resource.filePath.split('/').length === 2
```
topliceanurazvan, line 147: We should use `path.sep` instead of `/` because it is different on each OS system.

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -91,6 +129,41 @@ const NewResourceWizard = () => {
...
+      );
+    } else {
+      selectedFolderResources = Object.values(resourceMap).filter(
+        resource => resource.filePath.split('/').length > 2 && resource.filePath.startsWith(`/${selectedFolder}/`)
```
topliceanurazvan, line 151: Same here with `path.sep` instead of `/`. Also we should not check if the resource filePath starts with the selected folder as it will filter also all resources from all the subfolder from the selected folder, which is not the desired behaviour.

We could do something like this:

    Object.values(resourceMap).filter(resource => getDirname(resource.filePath).endsWith(selectedFolder));

Which will filter all resources whose parent folder ends with the selected folder.

topliceanurazvan (changes requested):
Left some comments

topliceanurazvan:
> @topliceanurazvan , have incorporated your suggestions. However, I see below edge cases,
> 
> 1. If resource name and/or kind is not defined, the export filename turns out to be `undefined`,
> 
> Screenshot 2022-06-20 at 9 05 06 PM
> 
> 2. Changing the form inputs (name, kind) is not triggering the changes
> 3. If conflicting resources are found and resource kind is not provided yet, the filename won’t be generated (`generateFullFileName` function fails @ [`NewResourceWizard.tsx#L52`](https://github.com/jenil777007/monokle/blob/6a4ba08c27bdc4bc52d81fb051315cba7a6e3779/src/components/organisms/NewResourceWizard/NewResourceWizard.tsx#L52))
> 4. Since resourceMap only contains the resources created via monokle, what if the file was created manually? Could raise a name conflict.

For `1.` and `3.` - completely agree, I would maybe disable the saving destination dropdown until the user enters the name/kind, @devcatalin wdyt?

For `2.` - we should definitely somehow trigger changes when the inputs are being changed, maybe on the onChange handler of the Form? 

For `4.` - Monokle watches your file system and if some file is being added manually, it will automatically add it in your `fileMap`, hence it will add corresponding resources to `resourceMap`.

devcatalin (changes requested):

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -27,12 +30,46 @@ import {openNamespaceTopic, openUniqueObjectNameTopic} from '@utils/shell';
...
+  resourceKind: K8sResource['kind'],
+  selectedFolder: string,
+  fileMap: FileMapType,
+  suffix: number,
+  includeKind?: boolean
+): string => {
+  const name = resourceName;
+  const nameKind = includeKind ? `-${resourceKind.toLowerCase()}` : '';
+  const nameSuffix = suffix ? ` (${suffix})` : '';
+  const fullFileName = fetchFullFileName(`${name}${nameKind}${nameSuffix}`);
```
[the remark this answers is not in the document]
jenil777007: That's right. Incorporated

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
jenil777007: Incorporated.

topliceanurazvan (changes requested):

devcatalin (approved):
LGTM

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -145,11 +145,12 @@ const NewResourceWizard = () => {
...
     if (selectedFolder === ROOT_FILE_ENTRY) {
       selectedFolderResources = Object.values(resourceMap).filter(
-        resource => resource.filePath.split('/').length === 2
+        resource => resource.filePath.split(path.sep).length === 2
       );
     } else {
       selectedFolderResources = Object.values(resourceMap).filter(
-        resource => resource.filePath.split('/').length > 2 && resource.filePath.startsWith(`/${selectedFolder}/`)
+        resource =>
+          resource.filePath.split(path.sep).length > 2 && getDirPath(resource.filePath).endsWith(selectedFolder)
```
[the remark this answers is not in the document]
jenil777007: Updated!

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -150,7 +146,7 @@ const NewResourceWizard = () => {
     } else {
       selectedFolderResources = Object.values(resourceMap).filter(
         resource =>
-          resource.filePath.split(path.sep).length > 2 && getDirPath(resource.filePath).endsWith(selectedFolder)
+          resource.filePath.split(path.sep).length > 2 && path.dirname(resource.filePath).endsWith(selectedFolder)
```
topliceanurazvan, line 149: We should use `path.dirname` depending on the os platform;

```
const osPlatform = useAppSelector(state => state.config.osPlatform);

const getDirname = osPlatform === 'win32' ? path.win32.dirname : path.dirname;.

...getDirname(resource.filePath).endsWith(selectedFolder);
```

topliceanurazvan (changes requested):

src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
jenil777007: My bad @topliceanurazvan I was modifying this only!

jenil777007: Please check now

jenil777007:
> Great work @jenil777007 🙌🏻

**Thank you Razvan. This was not possible without you.** 

Thank you for allowing me to contribute and mentoring me consistently. This was my first contribution in a major tool that is used by many. This was also my first Electron project(also, completely new to k8s). Learnt so many new tools, approaches and practices. Will be looking forward to contribute more.

Thank you @devcatalin for the reviews as well.

topliceanurazvan:
Thank you for contributing to Monokle! You did an awesome work! 💯 

Feel free to pick another issue from our list that you might find interesting and of course do not hesitate to contact us if you need any help!

Best of luck! 🤞🏻

PR #1961, opened 2022-06-17
fix: cluster resource delete

topliceanurazvan:
## Changes

- Use kube config selector for getting the path and context used for creating the kube client

## How to test it

- Delete a resource from the cluster directly

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


WitoDelnat (approved):
LGTM

PR #1975, opened 2022-06-21
feat: new file action in file explorer for folders

topliceanurazvan:
## Changes

- Action to create new file inside a folder

## Screenshots

New file action in context menu:

image

Create file modal:

image

After creating new file:

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/CreateFileFolderModal/CreateFileFolderModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+                    reject(new Error("This field can't be empty"));
+                  }
+
+                  if (prohibitedFirstSymbols.some(symbol => symbol === fileFolderNameValue[0])) {
+                    reject(
+                      new Error(`${uiState.type} name can't start with prohibited symbol - "${fileFolderNameValue[0]}"`)
+                    );
+                  }
+
+                  if (checkIfEntityExists(path.join(uiState.rootDir, path.sep, fileFolderNameValue))) {
```
devcatalin, line 107: I don't think `path.sep` is needed when using `path.join`.
Also, the `checkIfEntityExists` function name is not very intuitive, maybe it should be renamed to `doesPathExist` ?

src/utils/files.ts
```diff
@@ -143,14 +143,18 @@ export async function renameEntity(
...
 }
 
-export function createFolder(rootDir: string, folderName: string, callback: (args: CreateFolderCallback) => any) {
-  return fs.mkdir(`${rootDir}${path.sep}${folderName}`, err => callback({rootDir, folderName, err}));
+export function createFolder(rootDir: string, folderName: string, callback: (args: CreateFileFolderCallback) => any) {
+  return fs.mkdir(`${rootDir}${path.sep}${folderName}`, err => callback({rootDir, fileFolderName: folderName, err}));
+}
+
+export function createFile(rootDir: string, fileName: string, callback: (args: CreateFileFolderCallback) => any) {
+  return fs.open(`${rootDir}${path.sep}${fileName}`, 'wx', err => callback({rootDir, fileFolderName: fileName, err}));
```
devcatalin, line 157:
```suggestion
  return fs.open(path.join(rootDir, fileName), 'wx', err => callback({rootDir, fileFolderName: fileName, err}));
```

devcatalin (changes requested):

src/components/organisms/CreateFileFolderModal/CreateFileFolderModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+                    reject(new Error('A file or folder with this name already exists in this location'));
```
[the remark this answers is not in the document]
topliceanurazvan: Here is different because we are checking for name conflict, so it doesn't matter if it's a file or a folder.

PR #1993, opened 2022-06-27
fix: restore the previous selection history when preview fails

DzoQiEuoi:
This PR...

## Fixes

- Restores the previous selection history when previewing a helm chart or kustomize fails.

## How to test it

- Build up a selection history by moving between files or resources, make a change that will lead to a failure when launching a preview, then click preview. After the preview fails check that the selection history has been restored to its previous state.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan (approved):
LGTM!

PR #2025, opened 2022-06-29
feat: resources popover for images

topliceanurazvan:
## Changes

- Add popover with corresponding resources links for each image

## Screenshots

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/navsections/ImagesSectionBlueprint/ImageOutgoingResourcesPopover.tsx
+++ b/src/navsections/ImagesSectionBlueprint/ImageOutgoingResourcesPopover.tsx
@@ -20,7 +20,7 @@ const ImageOutgoingResourcesPopover: React.FC<IProps> = ({resourcesIds}) => {
   const selectedResourceId = useAppSelector(state => state.main.selectedResourceId);
 
   const refs = useMemo(() => {
-    return resourcesIds.reduce((currentRefs, id) => {
+    return resourcesIds.reduce((currentRefs: {[key: string]: ResourceRef[]}, id) => {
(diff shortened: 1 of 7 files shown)


src/navsections/ImagesSectionBlueprint/ImageOutgoingResourcesPopover.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+      const imageRefs = resourceRefs.filter(ref => ref.type === 'outgoing' && ref.target?.type === 'image');
+
+      if (imageRefs.length) {
+        currentRefs[id] = imageRefs;
+      }
+
+      return currentRefs;
+    }, {} as {[key: string]: ResourceRef[]});
+  }, [resourceMap, resourcesIds]);
```
devcatalin, line 39:
```suggestion
  const refs = useMemo(() => {
    return resourcesIds.reduce((currentRefs: {[key: string]: ResourceRef[]}, id) => {
      const resource = resourceMap[id];
      const resourceRefs = resource.refs;

      if (!resourceRefs) {
        return currentRefs;
      }

      const imageRefs = resourceRefs.filter(ref => ref.type === 'outgoing' && ref.target?.type === 'image');

      if (imageRefs.length) {
        currentRefs[id] = imageRefs;
      }

      return currentRefs;
    }, {});
  }, [resourceMap, resourcesIds]);
```

devcatalin (changes requested):

Issue #1606, opened 2022-04-14
issuerRef is shown as invalid when group property is missing 


olensmar:
see screenshot - the issuerRef in my Certificate is shown as invalid because the default value for the missing group property is not used to resolve the target resource 

<img width="997" alt="image" src="https://user-images.githubusercontent.com/1917063/163347623-fdd95f5a-ba95-4e60-91ee-d01e6c0f3bfb.png">

#1605 provides a possible solution

see https://cert-manager.io/docs/concepts/certificaterequest/ for more context

Issue #1605, opened 2022-04-14
Make siblingMatchers configurable


olensmar:
I need to specify a default target value for the groupMatcher - as this property is not required in many cases. 

See for example https://github.com/kubeshop/monokle/blob/main/resources/kindhandlers/handlers/cert-manager.io/CertificateRequest.json#L14 - this groupMatcher currently requires a group property in the corresponding issuerRef - but this property is not mandatory and defaults to the "cert-manager.io" group. I would like to specify this default value as a property of the matcher - instead of having to create a custom matcher for this specific case. 

A simple solution would be to add

```json
"matcherProperties" : {
   "groupMatcher" : {
      "defaultValue" : "cert-manager.io/v1"
   }
}
```

to the source configuration of the refMapper - the groupMatcher properties would be passed to the matcher on invocation.

PR #2038, opened 2022-07-01
feat: configurable sibling matchers

olensmar:
This PR adds the possibility to configure siblingMatchers (#1605) and fixes the corresponding cert-manager issue (#1606)

image

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [X] added a test


src/models/resourcekindhandler.ts
```diff
@@ -19,10 +19,14 @@ interface RefMapper {
...
         targetResource: K8sResource,
         value: string,
-        siblingValues: Record<string, string>
+        siblingValues: Record<string, string>,
+        properties?: any
       ) => boolean
     >;
 
+    // optional matcher configuration properties that will be passed to each correspondingly named matcher
+    matcherProperties?: Record<string, any>;
```
devcatalin, line 28: Maybe we should refactor this to:
```ts
type Matcher = 'kindMatcher' | 'groupMatcher' // ... add additional matcher types

import {JsonObject} from "type-fest"; // using this type to ensure RefMappers will remain serializable

interface RefMapper {
  // ...
  matcherProperties: Record<Matcher, JsonObject>
  // ...
}

devcatalin (changes requested):

src/models/resourcekindhandler.ts
olensmar: sure - I think this would be fine until we allow plugins to provide their own matchers... !?

olensmar: actually - the key of this map is dynamic - i.e. depends on the property to be matched - so we can't create a type with a fixed set of names. I'll add the JsonObject though 👍

devcatalin (approved):
LGTM

PR #2046, opened 2022-07-04
fix: mentioned the file path only once

topliceanurazvan:
## Fixes

- Mention the path of the found file only once 

## Screenshots

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


devcatalin (changes requested):
Let's change `Go to` to `Found in:` in that popup

devcatalin (approved):

PR #2049, opened 2022-07-05
feat: initial implementation of local server for hosting editor schemas

olensmar:
This PR starts a local server for hosting custom schemas requested by the editor. The server is started in a worker window - which could be expanded for further async-processing later on. 

The main (only?) drawback with this approach is that the user will be prompted to allow local network connections when the server starts the first time (which it does lazily as needed - which might be never for many users...).

## Checklist

- [X] tested locally
- [X] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/hooks/useResourceYamlSchema.ts
```diff
@@ -28,16 +33,30 @@ function useResourceYamlSchema(
...
+      let schemaUri = `https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/v${k8sVersion}/${resource?.kind.toLowerCase()}.json`;
+
+      // if this is a custom resource then host the schema using the local server
+      if (resourceKindHandler && resourceKindHandler.isCustom) {
+        const schemaPath = `/schemas/${resource?.kind.toLowerCase()}.json`;
+        ipcRenderer.send('add-local-server-content', {
+          path: schemaPath,
+          content: JSON.stringify(resourceSchema, null, 2),
+        });
+        schemaUri = `http://localhost:51038${schemaPath}`;
```
olensmar, line 57: port is hardcoded - need to get the actual port from the server

PR #2050, opened 2022-07-05
fix: reprocess helm value file/template

topliceanurazvan:
## Changes

-  Create Helm Templates map

## Fixes

- Reprocess helm template/value file after updating file entry

## Screenshots

Before modifying the file:

image

After modifying the file:

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/redux/services/helm.ts
+++ b/src/redux/services/helm.ts
@@ -169,11 +169,34 @@ export function createHelmTemplate(
     values: valueRanges,
   };
 
-  console.log('Helm template: ', helmTemplate);
-
   helmTemplatesMap[helmTemplate.id] = helmTemplate;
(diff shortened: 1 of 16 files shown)


src/redux/services/helm.ts
```diff
@@ -295,6 +332,30 @@ export const getObjectKeys = (obj: any, prefix = ''): string[] =>
     return [...res, prefix + el];
   }, []);
 
+export const getHelmValueFileValues = (fileContent: string) => {
```
devcatalin, line 335: maybe this should be `getHelmTemplateReferencedValues` ?

devcatalin (changes requested):

PR #2116, opened 2022-07-19
Julia/feat/edit text files

jmelnych:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/redux/initialState.ts
+++ b/src/redux/initialState.ts
@@ -186,7 +186,7 @@ const initialUiState: UiState = {
     expandedFolders: [],
     expandedSearchedFiles: ['filter'],
     isValidationDrawerVisible: false,
-    activeTab: undefined,
+    activeTab: null,
   },
   rightMenu: {
     isActive: electronStore.get('ui.rightMenu.isActive'),
--- a/src/redux/initialState.ts
+++ b/src/redux/initialState.ts
@@ -72,6 +72,7 @@ const initialAppState: AppState = {
     searchHistory: electronStore.get('appConfig.recentSearch') || [],
     currentMatch: null,
   },
+  supportedExtensions: ['.txt', '.md', '.html', '.json'],
 };
 
 const initialAppConfigState: AppConfig = {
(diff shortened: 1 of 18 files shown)


src/redux/initialState.ts
```diff
@@ -72,6 +72,7 @@ const initialAppState: AppState = {
     searchHistory: electronStore.get('appConfig.recentSearch') || [],
     currentMatch: null,
   },
+  supportedExtensions: ['.txt', '.md', '.html', '.json'],
```
jmelnych, line 75: these are the basic extensions I can see we might want to edit, please advice any other we can start with?

jmelnych, line 75: these are the basic extensions I can see we might want to edit, please advice any other we can start with?

jmelnych, line 75: these are the basic extensions I can see we might want to edit, please advice any other we can start with?

src/redux/selectors.ts
```diff
@@ -166,6 +166,13 @@ export const fileIncludesSelector = createSelector(
   }
 );
 
+export const supportedExtensionsSelector = createSelector(
```
topliceanurazvan, line 169: Do we need a selector if we are returning directly the `supportedExtensions` from the state? 🤔

topliceanurazvan (changes requested):
Looks good 👍🏻 

Noticed that the files that we are now supporting to edit are still marked as `unsupported` in the file explorer, even though we are able to click them. We should not mark them as unsupported anymore I think

image

image

src/redux/initialState.ts
devcatalin: we should include all text files extensions from here https://github.com/sindresorhus/text-extensions#readme

devcatalin (changes requested):

src/redux/selectors.ts
jmelnych: removed

topliceanurazvan (approved):
Great work 🙌🏻

devcatalin (approved):
LGTM and we can merge it, I'll follow-up on this next week to better define the `Files: Include` and `Files: Exclude` settings which we might not need anymore

PR #2132, opened 2022-07-25
feat: terminal window

topliceanurazvan:
## Changes

- Add terminal at the bottom
- Add left menu bottom options ( just the terminal for now )

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [x ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


electron/app/ipc/ipcListeners.ts
```diff
@@ -263,3 +269,75 @@ ipcMain.on('global-electron-store-update', (event, args: any) => {
     log.warn(`received invalid event type for global electron store update ${args.eventType}`);
   }
 });
+
+ipcMain.on('init-shell', (event, args) => {
```
devcatalin, line 273: let's write the event names in a similar format like you did below
```suggestion
ipcMain.on('shell.init', (event, args) => {
```

electron/app/ipc/ipcListeners.ts
```diff
@@ -263,3 +269,75 @@ ipcMain.on('global-electron-store-update', (event, args: any) => {
...
+      });
+    } else {
+      log.error('Web contents is not found');
+    }
+  } catch (e) {
+    log.error('Pty process could not be created ', e);
+  }
+});
+
+ipcMain.on('resize-shell', (event, args) => {
```
devcatalin, line 305:
```suggestion
ipcMain.on('shell.resize', (event, args) => {
```

package.json
```diff
@@ -133,6 +133,7 @@
     "new-github-issue-url": "1.0.0",
     "node-fetch": "2.6.7",
     "node-machine-id": "1.1.12",
+    "node-pty": "^0.10.1",
```
devcatalin, line 136: remove the `^` from the newly installed libs

src/components/organisms/PaneManager/PaneManager.tsx
```diff
@@ -23,14 +34,49 @@ const PaneManager: React.FC = () => {
     return 'max-content 1fr';
   }, [activeProject, isStartProjectPaneVisible]);
 
+  const handleBottomPaneResize = useCallback(
+    (elements: any) => {
+      const bottomPaneElement = elements.find((el: any) => el.props.id === 'bottomPane');
```
devcatalin, line 39: what if `bottomPaneElement` will be `undefined` ? maybe the case should be handled

src/models/ui.ts
```diff
@@ -59,6 +59,7 @@ export type LeftMenuSelectionType =
   | 'images-pane'
   | 'validation-pane'
   | 'search';
+export type LeftMenuBottomSelectionType = 'terminal' | null;
```
devcatalin, line 62: I think I would've rather typed this as `'terminal' | undefined` since we haven't really used `null` in the project, but if it's useful here we can leave it like that

src/redux/initialState.ts
```diff
@@ -72,6 +72,7 @@ const initialAppState: AppState = {
     searchHistory: electronStore.get('appConfig.recentSearch') || [],
     currentMatch: null,
   },
+  webContentsId: 1,
```
devcatalin, line 75: maybe this should be `undefined` by default? not sure how well it will work with multiple windows if we set it to `1`

devcatalin (changes requested):

src/models/ui.ts
topliceanurazvan: This is needed so we can deselect the options from the bottom menu, after one of them was selected already

devcatalin (approved):

PR #2195, opened 2022-08-09
Erdikose/feat/live resource updates

erdkse:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/redux/services/clusterResourceWatcher.ts
```diff
@@ -0,0 +1,119 @@
...
+        const [resource]: K8sResource[] = extractK8sResources(
+          jsonToYaml(apiObj),
+          PREVIEW_PREFIX + kubeConfig.currentContext
+        );
+        if (!previewResources[resource.id]) {
+          dispatch(updateClusterResource(resource));
+        }
+      }
+
+      if (type === 'MODIFIED') {
```
yasarkubeshop, line 86: looks like a good spot for switch(type)

src/redux/services/clusterResourceWatcher.ts
```diff
@@ -0,0 +1,119 @@
...
+        const [resource]: K8sResource[] = extractK8sResources(
+          jsonToYaml(apiObj),
+          PREVIEW_PREFIX + kubeConfig.currentContext
+        );
+
+        dispatch(updateClusterResource(resource));
+      }
+      if (type === 'DELETED') {
+        const [resource]: K8sResource[] = extractK8sResources(
```
yasarkubeshop, line 95: might worth to extract this 
```
const [resource]: K8sResource[] = extractK8sResources(
          jsonToYaml(apiObj),
          PREVIEW_PREFIX + kubeConfig.currentContext
        );
        dispatch(deleteClusterResource(resource));
```

src/redux/services/clusterResourceWatcher.ts
erdkse: I personally don't like switch-cases, I perefer if-return more

src/redux/services/clusterResourceWatcher.ts
erdkse: I thought the same too but I didn't extract it since it is a single line, But I think I will extract

src/redux/services/clusterResourceWatcher.ts
yasarkubeshop: Got it sure mate in that case lets switch to else if for the following ifs.

yasarkubeshop (approved):
utAck

src/redux/services/clusterResourceWatcher.ts
yasarkubeshop: Hey mate actually on second look we can even get it out from if blocks since they are same for each case

src/redux/services/clusterResourceWatcher.ts
erdkse: I added returns to all of them, In my opinion, writing `switch-case` or`elses` makes code hard to read, so I went with

devcatalin (dismissed):

yasarkubeshop (approved):

PR #2239, opened 2022-08-15
fix: cluster mode indicator in header for smaller screens; default he…

jmelnych:
…ight for bottom panel

This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots
Screenshot 2022-08-14 at 18 03 05

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan (changes requested):
There is still an overflow in the Actions pane header

image

topliceanurazvan (changes requested):
Can we make the minimum width for showing the dots smaller? Seems like there is some more space available.

Also do you think it's possible to trigger the dots menu while resizing and not after releasing the drag?

image

image

jmelnych:
> Can we make the minimum width for showing the dots smaller? Seems like there is some more space available.
> 
> Also do you think it's possible to trigger the dots menu while resizing and not after releasing the drag?
> 
> image
> 
> image

We can get editorPaneSize value only after resizing. In order to get trigger dots menu dynamically, I would need to use useWindowSize only but the pane sizes and how many of them open can vary, so useWindowSize does not act as a reliable item to base decision to hide or show action buttons.
Maybe you can suggest another strategy but this is what I'm thinking right now.

topliceanurazvan:
@jmelnych I was thinking to use `useMeasure` for the pane directly which should give us the width dynamically, wdyt?

topliceanurazvan (approved):
lgtm !! Great work 🙌🏻

Issue #2260, opened 2022-08-22
Monokle freezes when opening specific folders


chargio:
**Describe the bug**
Monokle freezes when trying to open the git clone for helm (helm/helm)

**To Reproduce**
1.  Clone helm helm into a local folder
2. Open Monokle
3. Point it to the local folder with helm
4. See it freeze

**Expected behavior**
It should open the folder without any problem

**Screenshots**
<img width="1199" alt="image" src="https://user-images.githubusercontent.com/7500590/185959057-0dd0d619-eae1-4460-b4c4-3fe686d97116.png">
<img width="1198" alt="image" src="https://user-images.githubusercontent.com/7500590/185959125-e6c23027-080c-4bbc-babc-02f37041d4f8.png">

**Desktop (please complete the following information):**
 - OS: Mac OS Monterey 12.5.1 (M1 Max)
Monokle 1.10.1

**Additional context**
Open Developer tools is not working, the console gets frozen too.

olensmar:
See #2263 - Monokle doesn't lock up - but it still takes a while to render due to rendering performance issues when there are 123 (!) Helm Charts.. waiting for @devcatalin

olensmar:
perhaps https://github.com/kubeshop/monokle/issues/1719 is related?

starcraft66:
This seems to still happen. Opening my infrastructure repo (https://github.com/starcraft66/infrastructure/) causes the app to freeze. I let it run for like 20 minutes and the UI eventually unblocked but it still didn't show anything more than what's visible in the issue's screenshot.

starcraft66:
I think that what is happening is that it is scanning my `.direnv` folder which contains symlinks to huge checkouts of `nixpkgs` and other repositories that are just meaningless noise. Is there a way to make it ignore folders? It finally loaded after a long time and it says it processed 34000 files in my repo when there are only few hundred relevant yaml files.

olensmar:
thanks for getting back to us @starcraft66 - yes, there is a setting for this under the global settings (as shown below) and then under project settings where they can be overridden... perhaps we should add `.direnv` to the default list of excludes?

<img width="1280" alt="image" src="https://github.com/kubeshop/monokle/assets/1917063/6d20095f-fafc-4812-86a5-d232db1dcee6">

olensmar:
another thought could be to use any found .gitignore files to further ignore - would that have helped in your case!?

starcraft66:
[...]

PR #2263, opened 2022-08-23
fix: improved handling of invalid Chart.yaml

olensmar:
This PR fixes handling of invalid Chart.yaml files - but still needs improvements in regard to rendering performance of the Helm panel when loading the helm/helm repo - see #2260 

@devcatalin can you look into the performance - you can perhaps use this branch for improving?


devcatalin:
Let's merge this as is, the performance will get fixed with the refactor of section blueprints for which I have a draft PR opened

devcatalin (approved):

PR #2307, opened 2022-08-30
feat: highlight last changed line

jmelnych:
This PR...

## Changes

-

## Fixes

-

## How to test it

1) select resource
2) go to form tab
3) make changes to some properties
4) go to code editor tab

expected result:
you should see the last modified prop highlighted in code editor.
Highlight disappears once a resource's code is modified or another resource is selected 

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/SearchPane/CreateFilteredNode.tsx
+++ b/src/components/organisms/SearchPane/CreateFilteredNode.tsx
@@ -39,7 +39,7 @@ export const createFilteredNode = (filteredFileMap: FileEntry[]): FilterTreeNode
           parentKey: item.filePath,
           title: (
             <Highlighter
-              highlightClassName="match-higlight"
+              highlightClassName="match-highlight"
               searchWords={line[0].matchesInLine}
               autoEscape
               textToHighlight={line[0].wholeLine}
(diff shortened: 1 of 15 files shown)


jmelnych:
[Julio]

So I think since it happens in the top of the page you easily can miss it. Can we take the focus around center of the page, instead on top? also, expected behaviour is that the line stays until you focus out of the editor or clicks on it to do some change. And that it would sort of blink at first

src/components/organisms/SearchPane/CreateFilteredNode.tsx
```diff
@@ -39,7 +39,7 @@ export const createFilteredNode = (filteredFileMap: FileEntry[]): FilterTreeNode
           parentKey: item.filePath,
           title: (
             <Highlighter
-              highlightClassName="match-higlight"
+              highlightClassName="match-highlight"
```
topliceanurazvan, line 42: Nice catch 👍🏻

topliceanurazvan (approved):
Great work, lgtm! 🙌🏻

Issue #2306, opened 2022-08-30
Added options for different terminal shell


dipankardas011:
**Is your feature request related to a problem? Please describe.**
In linux different shell options **unavailable**
![image](https://user-images.githubusercontent.com/65275144/187491258-d108463b-5d7f-4244-a19b-5668681cede6.png)

**Describe the solution you'd like**
To add entry to the array of terminals

dipankardas011:
![image](https://user-images.githubusercontent.com/65275144/187501629-b39a61b8-c8db-4b58-8ee9-6fda38792014.png)
making PR

PR #2308, opened 2022-08-30
feat(terminal): added ZSH shell for Linux based OS

dipankardas011:
Signed-off-by: Dipankar Das <dipankardas0115@gmail.com>

This PR...

## Changes

- Addition of ZSH shell for the 'other' OS

## Fixes

- Issue #2306 

## Screenshots

Screenshot from 2022-08-30 22-47-02

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/constants/terminal.ts
+++ b/src/constants/terminal.ts
@@ -1,5 +1,5 @@
 export const SUPPORTED_SHELLS = {
   win32: ['cmd.exe', 'powershell.exe'],
   darwin: ['bash', 'zsh'],
-  other: ['bash', 'zsh'],
+  other: ['bash', 'zsh', 'sh'],
 };
--- a/src/constants/terminal.ts
+++ b/src/constants/terminal.ts
@@ -1,5 +1,5 @@
 export const SUPPORTED_SHELLS = {
   win32: ['cmd.exe', 'powershell.exe'],
   darwin: ['bash', 'zsh'],
-  other: ['bash'],
+  other: ['bash', 'zsh'],
 };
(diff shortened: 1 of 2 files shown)


src/constants/terminal.ts
```diff
@@ -1,5 +1,5 @@
 export const SUPPORTED_SHELLS = {
```
topliceanurazvan, line 1: Let's make the SUPPORTED_SHELLS a record of `<string, string[]>`

src/redux/services/terminalShells.ts
```diff
@@ -28,6 +28,7 @@ const detectAvailableUnixShells = (osPlatform: NodeJS.Platform): ShellsMapType =
     });
   } else {
```
topliceanurazvan, line 29: Here we should check the same as for the `darwin` platform and only add the `zsh` and `bash` if they are found in `/etc/shells` folder

topliceanurazvan (changes requested):
Hello @dipankardas011 and thank you for contributing to Monokle! 🙌🏻 

I've left some comments regarding your PR, if you have any questions please let us know.

src/constants/terminal.ts
dipankardas011: ig it's is in this form
image

src/redux/services/terminalShells.ts
dipankardas011: Got it going 
image

src/constants/terminal.ts
dipankardas011: ig its already resolved just added some more shells which many people use

dipankardas011:
@topliceanurazvan changed as you suggested

topliceanurazvan (approved):
Great work! 🙌🏻

dipankardas011:
Thanks for your review

chargio:
Great!

PR #2329, opened 2022-09-09
feat: Julia/feat/code form editor

jmelnych:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

image

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


devcatalin:
image

I'm getting an error when I select the `Form` tab on a resource

topliceanurazvan (changes requested):
Code looks good, great work 🙌🏻  Found just a little styling problem with some missing padding.

image

Also I'm thinking if we should not have all the levels directly expanded as it hits the performance a little bit on rendering all of them - for big forms especially - wdyt? @jmelnych @devcatalin

topliceanurazvan (changes requested):
We should have same px for both left and right paddings ( 10px )

image

image

topliceanurazvan (approved):

PR #2508, opened 2022-10-07
refactor: git clone modal

topliceanurazvan:
## Changes

- Clone git repo in a new folder named after repo ( throwing error if already exists ), inside the selected location.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


electron/app/git/git.ts
```diff
@@ -62,11 +62,11 @@ export async function cloneGitRepo(payload: {localPath: string; repoPath: string
   try {
     const stat = await fs.stat(localPath);
     if (!stat.isDirectory()) {
-      throw new Error(`${localPath} is ot a directory`);
+      throw new Error(`${localPath} is not a directory`);
     }
   } catch (e: any) {
     if (e.code === 'ENOENT') {
-      fs.mkdir(localPath);
+      await fs.mkdir(localPath);
```
devcatalin, line 69: Can you test this by entering a path where a folder and then the subfolder do not exist?
For example you have `C:\Users\Someone\Projects` that exists and then in the input you enter `C:\Users\Someone\Projects\Folder\Subfolder` where both Folder and Subfolder do not exist.
I think in this case we should show an error? 🤔

devcatalin (approved):

PR #2572, opened 2022-10-14
fix: dispose only models not attached to any editor

topliceanurazvan:
## Fixes

- Dispose only models that are not attached to any editor

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


devcatalin (approved):
LGTM

PR #2581, opened 2022-10-16
feat: Remove Metadata tab, bring back to Form tab

erdkse:
This PR...

## Changes

- Removes metadata tab and adds into form tab

## Fixes

-

## How to test it

-

## Screenshots

Screen Shot 2022-10-16 at 23 40 03

Screen Shot 2022-10-16 at 23 40 59

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/kindhandlers/common/formLoader.ts
+++ b/src/kindhandlers/common/formLoader.ts
@@ -8,9 +8,14 @@ const uiformSchemaCache = new Map<string, any>();
 export function getFormSchema(kind: string) {
   try {
     if (!formSchemaCache.has(kind)) {
+      const metadataSchema = loadResource(`form-schemas${path.sep}metadata-schema.json`);
       const formSchema = loadResource(`form-schemas${path.sep}${kind.toLowerCase()}-schema.json`);
-      if (formSchema) {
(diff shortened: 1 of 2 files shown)


src/kindhandlers/common/formLoader.ts
```diff
@@ -8,9 +8,14 @@ const uiformSchemaCache = new Map<string, any>();
 export function getFormSchema(kind: string) {
   try {
     if (!formSchemaCache.has(kind)) {
+      const metadataSchema = loadResource(`form-schemas${path.sep}metadata-schema.json`);
       const formSchema = loadResource(`form-schemas${path.sep}${kind.toLowerCase()}-schema.json`);
-      if (formSchema) {
-        formSchemaCache.set(kind, JSON.parse(formSchema));
+      if (formSchema && metadataSchema) {
```
topliceanurazvan, line 13: Can there be cases where the Form schema is available, but the metada not? 🤔

erdkse: I know as all resources should have metadata. I don't know a case that it doesn't have

topliceanurazvan (approved):

PR #2666, opened 2022-11-09
feat(menu): open on github

yasarkubeshop:
This PR...

## Changes

Menu on file explorer

## Fixes

-

## How to test it

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/electron/app/git/git.ts
+++ b/electron/app/git/git.ts
@@ -23,6 +23,12 @@ export async function isGitInstalled(path: string) {
   }
 }
 
+export async function getGitRemoteUrl(path: string) {
+  const git: SimpleGit = simpleGit({baseDir: path});
+  const result = await git.raw('config', '--get','remote.origin.url');
+  return result;
+}
+
 export async function areFoldersGitRepos(paths: string[]) {
   let foldersStatus: {path: string; isGitRepo: boolean}[] = [];
 
--- a/electron/app/git/git.ts
+++ b/electron/app/git/git.ts
@@ -105,8 +105,8 @@ export async function getGitRepoInfo(localPath: string) {
       hasRemoteRepo: false,
     };
 
-    if(remoteUrl!==""){
-      gitRepo.remoteUrl=remoteUrl;
+    if(remoteUrl){
+      gitRepo.remoteUrl = remoteUrl;
     }
 
     gitRepo.branchMap = Object.fromEntries(
--- a/electron/app/git/git.ts
+++ b/electron/app/git/git.ts
@@ -103,8 +103,11 @@ export async function getGitRepoInfo(localPath: string) {
       branchMap: {},
       commits: {ahead: 0, behind: 0},
       hasRemoteRepo: false,
-      remoteUrl,
     };
+
+    if(remoteUrl!==""){
+      gitRepo.remoteUrl=remoteUrl;
+    }
 
     gitRepo.branchMap = Object.fromEntries(
       Object.entries({...localBranches.branches}).map(([key, value]) => [
--- a/electron/app/git/git.ts
+++ b/electron/app/git/git.ts
@@ -95,6 +95,7 @@ export async function getGitRepoInfo(localPath: string) {
   try {
     const remoteBranchSummary = await git.branch({'-r': null});
     const localBranches = await git.branchLocal();
+    const remoteUrl = await getGitRemoteUrl(localPath);
 
     gitRepo = {
       branches: [...localBranches.all, ...remoteBranchSummary.all],
@@ -102,6 +103,7 @@ export async function getGitRepoInfo(localPath: string) {
       branchMap: {},
       commits: {ahead: 0, behind: 0},
       hasRemoteRepo: false,
+      remoteUrl,
     };
 
     gitRepo.branchMap = Object.fromEntries(
--- a/src/components/organisms/FileTreePane/TreeItem.tsx
+++ b/src/components/organisms/FileTreePane/TreeItem.tsx
@@ -24,7 +24,6 @@ import {Dots, Spinner} from '@atoms';
 
 import {defineHotkey} from '@utils/defineHotkey';
 import {deleteEntity} from '@utils/files';
-import {promiseFromIpcRenderer} from '@utils/promises';
 import {showItemInFolder} from '@utils/shell';
 
 import Colors from '@styles/Colors';
@@ -73,7 +72,6 @@ const TreeItem: React.FC<TreeItemProps> = props => {
   } = props;
 
   const [isTitleHovered, setTitleHoverState] = useState(false);
-  const [remotePathUrl, setRemotePathUrl] = useState<undefined | string>();
 
   const fileOrFolderContainedInFilter = useAppSelector(state => state.main.resourceFilter.fileOrFolderContainedIn);
   const fileMap = useAppSelector(state => state.main.fileMap);
@@ -82,14 +80,6 @@ const TreeItem: React.FC<TreeItemProps> = props => {
   const resourceMap = useAppSelector(state => state.main.resourceMap);
   const helmValuesMap = useAppSelector(state => state.main.helmValuesMap);
   const git = useAppSelector(state => state.git);
-  const selectedProjectRootFolder = useAppSelector(state => state.config.selectedProjectRootFolder);
-  const remotePath = useMemo(
-    async () =>
-      selectedProjectRootFolder
-        ? promiseFromIpcRenderer('git.getGitRemoteUrl', 'git.getGitRemoteUrl.result', selectedProjectRootFolder)
-        : null,
-    [selectedProjectRootFolder]
-  );
 
   const isInPreviewMode = useSelector(isInPreviewModeSelector);
 
@@ -125,19 +115,6 @@ const TreeItem: React.FC<TreeItemProps> = props => {
 
     return style;
   }, [isMatchItem]);
-
-  React.useEffect(() => {
-    let mounted = true;
-    (async () => {
-      const res = await remotePath;
-      if (mounted) {
-        setRemotePathUrl(res);
-      }
-    })();
-    return () => {
-      mounted = false;
-    };
-  }, [remotePath]);
 
   useHotkeys(
     defineHotkey(hotkeys.DELETE_RESOURCE.key),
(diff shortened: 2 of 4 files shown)


electron/app/git/git.ts
```diff
@@ -23,6 +23,12 @@ export async function isGitInstalled(path: string) {
   }
 }
 
+export async function getGitTopLevel(path: string) {
```
topliceanurazvan, line 26: Maybe something more specific here could be `getGitRemoteUrl` ?

electron/app/git/git.ts
```diff
@@ -51,6 +57,7 @@ export async function isFolderGitRepo(path: string) {
 }
 
 export async function getRemotePath(localPath: string) {
+  console.log("loca"+localPath);
```
topliceanurazvan, line 60: I think we can remove this

src/components/organisms/FileTreePane/TreeItem.tsx
```diff
@@ -71,12 +74,19 @@ const TreeItem: React.FC<TreeItemProps> = props => {
...
+
   const fileOrFolderContainedInFilter = useAppSelector(state => state.main.resourceFilter.fileOrFolderContainedIn);
   const fileMap = useAppSelector(state => state.main.fileMap);
   const osPlatform = useAppSelector(state => state.config.osPlatform);
   const selectedPath = useAppSelector(state => state.main.selectedPath);
   const resourceMap = useAppSelector(state => state.main.resourceMap);
   const helmValuesMap = useAppSelector(state => state.main.helmValuesMap);
+  const git = useAppSelector(state => state.git);
+  const selectedProjectRootFolder = useAppSelector(state => state.config.selectedProjectRootFolder);
+  const remotePath = async () => (selectedProjectRootFolder ? getGitTopLevel(selectedProjectRootFolder) : null);
```
topliceanurazvan, line 89: Can we maybe switch it to a `useMemo` and use promiseFromIpcRenderer directly inside? In this way, we don't have to wait for creating the remote path ( when clicking on the open on GitHub option ), and only after that open the link

src/components/organisms/FileTreePane/TreeItem.tsx
```diff
@@ -301,6 +311,19 @@ const TreeItem: React.FC<TreeItemProps> = props => {
         ]
       : []),
     {key: 'divider-4', type: 'divider'},
+    ...(remotePath
```
topliceanurazvan, line 314: We should check if the git repo exists and if the remote is a GitHub repo, otherwise, the option should not be available ( hidden )

topliceanurazvan (changes requested):

src/components/organisms/FileTreePane/TreeItem.tsx
yasarkubeshop: I think this ...( syntax doing it if remote path is there its an none empty array and ifts not it will not show anything

topliceanurazvan: But what if the `remotePath` is a GitLab one?

yasarkubeshop: oh you are right I need to handle inside getGitTopLevel function, not sure if it returns null in case its not github will test this 👍

yasarkubeshop: Pushed an update for this 👍

src/components/organisms/FileTreePane/TreeItem.tsx
```diff
@@ -124,6 +126,19 @@ const TreeItem: React.FC<TreeItemProps> = props => {
     return style;
   }, [isMatchItem]);
 
+  React.useEffect(() => {
+    let mounted = true;
```
topliceanurazvan, line 130: Do we really need to check for mounted?

src/components/organisms/FileTreePane/TreeItem.tsx
```diff
@@ -124,6 +126,19 @@ const TreeItem: React.FC<TreeItemProps> = props => {
...
+    (async () => {
+      const res = await remotePath;
```
topliceanurazvan, line 132: Maybe we could just use the useEffect directly for setting the `remotePathUrl`, without the need for the useMemo. Or is there any particular reason for using both?

src/components/organisms/FileTreePane/TreeItem.tsx
```diff
@@ -313,15 +328,14 @@ const TreeItem: React.FC<TreeItemProps> = props => {
         ]
       : []),
     {key: 'divider-4', type: 'divider'},
-    ...(remotePath
+    ...(remotePathUrl && remotePathUrl.includes('github')
```
topliceanurazvan, line 331: I think a better check here would be for `https://github.com` - there may be cases where for example in a gitlab repo the name to be `my-github-test` ...

topliceanurazvan (changes requested):

src/components/organisms/FileTreePane/TreeItem.tsx
yasarkubeshop: Hey mate yes I saw the memory leak when program starts we need to check this

src/components/organisms/FileTreePane/TreeItem.tsx
yasarkubeshop: oh absolutely thanks mate 👍

src/components/organisms/FileTreePane/TreeItem.tsx
yasarkubeshop: Hey mate thought of this too but we are making a concat there when user clicks on open on github, thought maybe we dont want to do that unless the link is github and if user doesnt click.

src/components/organisms/FileTreePane/TreeItem.tsx
```diff
@@ -70,13 +73,24 @@ const TreeItem: React.FC<TreeItemProps> = props => {
...
   const osPlatform = useAppSelector(state => state.config.osPlatform);
   const selectedPath = useAppSelector(state => state.main.selectedPath);
   const resourceMap = useAppSelector(state => state.main.resourceMap);
   const helmValuesMap = useAppSelector(state => state.main.helmValuesMap);
+  const git = useAppSelector(state => state.git);
+  const selectedProjectRootFolder = useAppSelector(state => state.config.selectedProjectRootFolder);
+  const remotePath = useMemo(
+    async () =>
+      selectedProjectRootFolder
+        ? promiseFromIpcRenderer('git.getGitRemoteUrl', 'git.getGitRemoteUrl.result', selectedProjectRootFolder)
```
devcatalin, line 89: I would like the remote to be stored directly in the redux store on the `GitRepo` model
In the `electron/app/git.ts` file you'll find the `getGitRepoInfo` method where the `gitRepo` object is initialized.
There, you could call the `getGitRemoteUrl` method that you've created and set the remote to the `gitRepo` object.
You'll have to add the new property on the GitRepo type as well.

src/components/organisms/FileTreePane/TreeItem.tsx
```diff
@@ -112,6 +126,19 @@ const TreeItem: React.FC<TreeItemProps> = props => {
     return style;
   }, [isMatchItem]);
 
+  React.useEffect(() => {
```
devcatalin, line 129: This hook will not be needed if we save the remote to the redux state, see my other comment.

devcatalin (changes requested):

src/components/organisms/FileTreePane/TreeItem.tsx
yasarkubeshop: yep feels cleaner thanks mate 👍

yasarkubeshop: updated like this 👍

src/components/organisms/FileTreePane/TreeItem.tsx
```diff
@@ -301,6 +305,18 @@ const TreeItem: React.FC<TreeItemProps> = props => {
         ]
       : []),
     {key: 'divider-4', type: 'divider'},
+    ...(git.repo
```
topliceanurazvan, line 308: We should still check in here if the `remoteURL` contains the `https://github.com`

topliceanurazvan (changes requested):

src/components/organisms/FileTreePane/TreeItem.tsx
devcatalin: Maybe check only for `github` ? Probably there are multiple domains that work..

topliceanurazvan: As mentioned before, what happens if your GitLab repo has a name like `my-github-test` ?

yasarkubeshop: Hey I thought if its not a github then repo will return falsy and I tried with gitlab it was not showing link but let me be sure 👍 I will check in detail and come back

yasarkubeshop: Hmm directly works on gitlab too since its a git repo too lol I am a little confused, should I put open on gitlab if its on gitlab? or not show anything? @devcatalin @topliceanurazvan

yasarkubeshop: updated to check only github for now as mentioned on task 👍

electron/app/git/git.ts
```diff
@@ -98,6 +105,10 @@ export async function getGitRepoInfo(localPath: string) {
       hasRemoteRepo: false,
     };
 
+    if(remoteUrl!==""){
```
topliceanurazvan, line 108:
```suggestion
    if(remoteUrl)
```

electron/app/git/git.ts
```diff
@@ -98,6 +105,10 @@ export async function getGitRepoInfo(localPath: string) {
...
+      gitRepo.remoteUrl=remoteUrl;
```
topliceanurazvan, line 109:
```suggestion
      gitRepo.remoteUrl = remoteUrl;
```

topliceanurazvan (changes requested):

topliceanurazvan (approved):

topliceanurazvan:
Great work 🙌🏻 !

devcatalin (approved):

Issue #2756, opened 2022-11-17
Redesign the header of the Editor


devcatalin:
The Header is in the `ActionsPane` component.
The new `TitleBar` component from 2.0 branch should be used.

Old design:
<img width="830" alt="image" src="https://user-images.githubusercontent.com/20538711/202437686-410ee32e-761a-42ae-ba10-4757b3099cf3.png">

New design:
<img width="1081" alt="image" src="https://user-images.githubusercontent.com/20538711/202437807-d58b08d1-722f-41e5-8bce-24539d8aad28.png">

PR #2768, opened 2022-11-20
feat(editor): design updates on editor header

yasarkubeshop:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/atoms/EditorMonoPaneTitle/EditorMonoPaneTitle.tsx
+++ b/src/components/atoms/EditorMonoPaneTitle/EditorMonoPaneTitle.tsx
@@ -1,0 +1,27 @@
+import {Typography} from 'antd';
+
+import styled from 'styled-components';
+
+import {FontColors} from '@styles/Colors';
+
+const {Text} = Typography;
+
+const EditorMonoPaneTitle = styled(Text)`
+  &.ant-typography {
+    white-space: nowrap;
+    display: flex;
+    justify-content: space-between;
+    margin-bottom: 0;
+    padding: 8px 16px 8px 16px;
+    font-size: 14px;
+    font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, 'Noto Sans', sans-serif,
+      'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
+    font-variant: tabular-nums;
+    font-style: normal;
+    font-weight: 600;
+    line-height: 24px;
+    color: ${FontColors.darkThemeMainFont};
+  }
+`;
+
+export default EditorMonoPaneTitle;
--- a/src/components/molecules/EditorTitleBar/EditorTitleBar.tsx
+++ b/src/components/molecules/EditorTitleBar/EditorTitleBar.tsx
@@ -1,0 +1,43 @@
+import React from 'react';
+
+import {Button} from 'antd';
+
+import {useAppDispatch} from '@redux/hooks';
+import {toggleLeftMenu} from '@redux/reducers/ui';
+
+import {EditorMonoPaneTitle} from '@atoms';
+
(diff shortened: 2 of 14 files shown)


src/components/atoms/EditorMonoPaneTitle/EditorMonoPaneTitle.tsx
```diff
@@ -0,0 +1,27 @@
+import {Typography} from 'antd';
+
+import styled from 'styled-components';
+
+import {FontColors} from '@styles/Colors';
+
+const {Text} = Typography;
+
+const EditorMonoPaneTitle = styled(Text)`
```
topliceanurazvan, line 9: Is this needed?

src/components/molecules/EditorTitleBar/EditorTitleBar.tsx
```diff
@@ -0,0 +1,43 @@
...
+import * as S from './EditorTitleBar.styled';
+
+interface IProps {
+  title: string;
+  leftButtons?: React.ReactNode;
+  closable?: boolean;
+  children?: React.ReactNode;
+}
+
+const EditorTitleBar: React.FC<IProps> = props => {
```
topliceanurazvan, line 19: We should use the `TitleBar` from inside `@monokle-core/components` instead of creating our own

src/components/organisms/ActionsPane/Diff/Diff.tsx
```diff
@@ -17,14 +17,14 @@ const Diff: React.FC<IProps> = props => {
 
   return (
     <Tooltip mouseEnterDelay={TOOLTIP_DELAY} title={tooltipTitle} placement="bottomLeft">
-      <PrimaryButton
+      <LinkButton
```
topliceanurazvan, line 20: Kind of confusing to be able to give a link button the type `primary`, it should only have the type of link

topliceanurazvan (changes requested):

src/components/atoms/EditorMonoPaneTitle/EditorMonoPaneTitle.tsx
yasarkubeshop: The same reason as below comment created new to not affect all others.

src/components/organisms/ActionsPane/Diff/Diff.tsx
yasarkubeshop: Thanks for the comment mate, I need a new button with no borders what should be the name? I used LinkButton since I saw it is used for such components before but open to alternative names.

topliceanurazvan: Can we directly use the antd button with type link without needing to create an alternate custom component?

src/components/molecules/EditorTitleBar/EditorTitleBar.tsx
topliceanurazvan: Why would the `TitleBar` need to be updated? It should already be implemented to support all needs, or is there something missing?

yasarkubeshop: It looks brand new I dont suppose it can support these changes mate because I added some new classes on tab that is not used before and without those can not update new tab styles I want to share the new look but I could not find the ticket 2756

yasarkubeshop: Found it :) https://github.com/kubeshop/monokle/issues/2756

yasarkubeshop:
@topliceanurazvan lets resolve issues on this mate then I will raise another pr with those changes to 2.0 if we change remote on this it becomes 43 files since source branch is different :)

yasarkubeshop:
Closing this since we moved this mr from 2.0 on here https://github.com/kubeshop/monokle/pull/2785/files

PR #2795, opened 2022-11-23
feat: add option to generate random content for new resources

olensmar:
added the following checkbox to the New Resource dialog (checking it disables the template selection):

image

Random content is generated - although I'm not sure it's always so helpful...

image

## Checklist

- [X] tested locally
- [X] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/NewResourceWizard/NewResourceWizard.tsx
```diff
@@ -567,12 +598,18 @@ const NewResourceWizard = () => {
...
+          </Checkbox>
+        </Form.Item>
+
         <Form.Item
           name="selectedResourceId"
           label="Select existing resource as template"
           initialValue={SELECT_OPTION_NONE}
         >
-          <Select showSearch>
+          <Select showSearch disabled={generateRandom === true}>
```
topliceanurazvan, line 612:
```suggestion
          <Select showSearch disabled={generateRandom}>
```

src/redux/services/unsavedResource.ts
```diff
@@ -35,16 +35,30 @@ export function createUnsavedResource(
...
-      ...jsonTemplate,
-      apiVersion: input.apiVersion,
-      kind: input.kind,
-      metadata: {
-        ...(jsonTemplate.metadata || {}),
-        name: input.name,
-        namespace: input.namespace,
-      },
-    };
+    if (jsonTemplate.kind && jsonTemplate.apiVersion) {
```
topliceanurazvan, line 38: We could use directly this here so we don't duplicate the object:

```
newResourceContent = {
        ...jsonTemplate,
        apiVersion: input.apiVersion,
        kind: input.kind,
        metadata: {
          ...(jsonTemplate.metadata || {}),
          name: input.name,
          namespace: input.namespace,
        },
        ...(jsonTemplate.kind && jsonTemplate.apiVersion && {...jsonTemplate}),
      };
```

topliceanurazvan (commented):
Looks good 👍🏻 left a couple of comments...

src/redux/services/unsavedResource.ts
olensmar: the problem I had with putting jsonTemplate first was that the apiVersion/kind/metadata properties were added to the end of the document - which was weird in the editor

topliceanurazvan (approved):

PR #2814, opened 2022-11-26
refactor: Resource & Selection models + redux state

devcatalin:
This PR...

## Goals
- remove validation & refs processing logic (which will be replaced by the `@monokle/validation` package)
- decouple UI state and content state of resources by splitting resources into `ResourceMeta` and `ResourceContent`
  - keeping a `K8sResource` model for easier refactoring, which will have to be merged/unmerged from/to meta and content
- separate the storage of resources into different maps
  - this introduces the `Origin` concept and removes file related properties from the resource
- unify selections into a single object & model
- using `runtypes` for defining the new models to easily generate type guards that will help working with each different type of resource

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/shared/models/origin.ts
```diff
@@ -0,0 +1,60 @@
...
+export type ClusterOrigin = {
+  type: 'cluster';
+  context: string;
+};
+export type HelmOrigin = {
+  type: 'helm';
+  chartId: string;
+  valuesFilePath: string;
+};
+export type KustomizeOrigin = {
```
olensmar, line 16: should we add a "CommandOrigin" for resources that are generated by a preview command?

devcatalin: good point, thanks! I'll add it

src/shared/models/k8sResource.ts
```diff
@@ -1,36 +1,24 @@
...
-type ResourceValidation = {
-  isValid: boolean;
-  errors: ResourceValidationError[];
-};
-
-/**
- * A k8s resource manifest, either extracted from a file or generated internally (for example when previewing kustomizations or helm charts)
- */
-type K8sResource = {
+export type ResourceMeta<Origin extends AnyOrigin = AnyOrigin> = {
```
WitoDelnat, line 17: I think it's easier to create a type union which you can discriminate based on a property instead of working with generics. You can look at compare ResourceSet as I use the pattern there as well.

```
type Origin = LocalOrigin | ClusterOrigin | ..;

type BaseOrigin = {
  type: string;
  // other shared properties
};

type LocalOrigin = BaseOrigin & {
  type: "local",
  // specific properties
  directory: string;
}

// Now you can easily discriminate
if (resource.origin.type === "local") {
  resource.origin.directory // correctly infers.
}
```

I guess you use it here as you want to enforce that it always has the same origin? Asserting it from AnyOrigin to SpecificOrigin is a bit more tedious though so wanted to give an alternative approach.

src/shared/models/k8sResource.ts
```diff
@@ -1,36 +1,24 @@
-import {Scalar} from 'yaml';
+import * as Rt from 'runtypes';
```
WitoDelnat, line 1: is runtypes new? I'd recommend Zod as my go to schema validator as it works tremendously well with TypeScript.

src/shared/models/origin.ts
```diff
@@ -0,0 +1,60 @@
+import * as Rt from 'runtypes';
+
+export type LocalOrigin = {
+  type: 'local';
+  filePath: string;
```
WitoDelnat, line 5: You will need `fileOffset` here to determine the error location in the file within monokle/validation.

src/shared/models/k8sResource.ts
devcatalin: The initial version of this wasn't using generics but I was thinking there will be too many different types to remember and then when I switched to the generic it seemed much easier to read

devcatalin:
Closing this PR in favor of https://github.com/kubeshop/monokle/pull/2999

PR #2908, opened 2022-12-15
fix: rename helm value file

topliceanurazvan:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/navsections/HelmChartSectionBlueprint/HelmChartContextMenu.tsx
+++ b/src/navsections/HelmChartSectionBlueprint/HelmChartContextMenu.tsx
@@ -9,6 +9,7 @@ import styled from 'styled-components';
 
 import {ROOT_FILE_ENTRY} from '@constants/constants';
 
+import {HelmValuesFile} from '@models/helm';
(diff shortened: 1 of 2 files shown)


src/navsections/HelmChartSectionBlueprint/HelmChartContextMenu.tsx
```diff
@@ -36,6 +37,16 @@ const StyledActionsMenuIconContainer = styled.span<{isSelected: boolean}>`
   align-items: center;
 `;
 
+// temporary solution for renaming value file
```
devcatalin, line 40: let's add the `// TODO: ` prefix to this comment and maybe an explanation of what the problem might be, so we have a starting point to tackle it later

devcatalin (changes requested):

devcatalin (approved):

PR #2910, opened 2022-12-16
Erdikose/feat/kubectl proxy

erdkse:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/utils/commands/execute.ts
```diff
@@ -40,6 +41,20 @@ export function runCommandInMainThread(options: CommandOptions): Promise<Command
...
+  log.info('sending command stream to main thread', options);
+  const commandEventEmitter = new events.EventEmitter();
+  commandEventEmitter.addListener(options.commandId, listener);
+
+  const cb = (_event: unknown, arg: any) => {
+    if (arg.commandId !== options.commandId) return;
+    commandEventEmitter.emit(options.commandId, arg);
+  };
+  ipcRenderer.on('command-stream-event', cb);
+  ipcRenderer.send('run-command-stream', options);
```
devcatalin, line 55: We should probably generate a unique id in this function and pass it to the `run-command-stream` via options.
Then, when we receive `command-stream-event` we check if the ID matches with the one we sent.
Otherwise I'm thinking that multiple calls to `runCommandStreamInMainThread` in different places could receive wrong data.

src/utils/kubeclient.ts
```diff
@@ -137,6 +139,38 @@ export const getKubeAccess = async (namespace: string, currentContext: string):
...
+      args: ['proxy', '--port', `${port}`],
+    });
+  }
+};
+
+export const killKubeProxy = (listener: (...args: any[]) => void) => {
+  if (isRendererThread()) {
+    runCommandStreamInMainThread(listener, {
+      commandId: uuid(),
+      cmd: 'pkill',
```
devcatalin, line 162: will this work on Windows as well?

devcatalin (changes requested):

src/utils/kubeclient.ts
erdkse: nope, I commented a comment on slack, but couldn't try it yet

src/utils/commands/execute.ts
devcatalin: I think you're correct, resolving this comment

topliceanurazvan (approved):

erdkse:
@devcatalin can you approve this too, It is blocked because you wanted changes

devcatalin (approved):

PR #2996, opened 2023-01-12
Mm/fix/k8sschemadropdown

mortada-codes:
This PR...

## Changes

- adding k8s schema version dropdown

- adding OPA rules count chip

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/PageHeader/K8sVersionSelection.tsx
```diff
@@ -0,0 +1,67 @@
...
+import {K8S_VERSIONS} from '@shared/constants';
+
+import * as S from './K8sVersionSelection.style';
+
+export const K8sVersionSelection = () => {
+  const dispatch = useAppDispatch();
+  const appConfig = useAppSelector(state => state.config);
+  const userDataDir = useAppSelector(state => state.config.userDataDir);
+
+  const selectedK8SVersion = String(appConfig?.k8sVersion);
```
devcatalin, line 18: I recommend the following change here because if we select the entire `state.config` object, then this component will re-render each time any of it's properties change. However, if we select the `k8sVersion` directly, then the component will re-render only when that changes.

```suggestion
  const selectedK8SVersion = useAppSelector(state => String(state.config.k8sVersion));
  const userDataDir = useAppSelector(state => state.config.userDataDir);
```

src/components/organisms/PageHeader/OPAChip.tsx
```diff
@@ -0,0 +1,15 @@
+import {useAppSelector} from '@redux/hooks';
+
+import {OPAStatus} from '@components/atoms/Icon/Icons';
+
+import * as S from './OPAChip.styled';
+
+export const OPAChip = () => {
+  const opaRulesCount = useAppSelector(state => state.validation.config.rules)?.length || 0;
```
devcatalin, line 8: The `.length` could go into the selector, it will also help avoiding re-renders if the number doesn't change but the rules themselves change.
```suggestion
  const opaRulesCount = useAppSelector(state => state.validation.config.rules.length || 0);
```

src/redux/thunks/downloadK8sSchema.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import log from 'loglevel';
+import path from 'path';
+
+import {updateK8sVersion} from '@redux/reducers/appConfig';
+import {downloadSchema} from '@redux/services/k8sVersionService';
+
+import {AppDispatch, RootState} from '@shared/models';
+
+export const downloadK8sSchema = createAsyncThunk<void, string, {dispatch: AppDispatch; state: RootState}>(
```
devcatalin, line 11: I think this has the logic from the `handleDownloadVersionSchema` method in the `Settings` component.
Could we use this thunk in the Settings component as well to reduce duplicate code?

src/utils/index.ts
```diff
@@ -17,3 +19,7 @@ export const getRelativeDate = (isoDate: string | undefined) => {
   }
   return '';
 };
+
+export const doesSchemaExist = (k8sVersion: string, userDataDir?: string) => {
```
devcatalin, line 23: any reason for having `userDataDir` as optional here?

devcatalin (changes requested):

src/components/organisms_new/SettingsPane/Settings/Settings.tsx
```diff
@@ -271,10 +269,8 @@ export const Settings = ({
     }
   };
 
-  const doesSchemaExist = useCallback(
-    (k8sVersion: string) => {
-      return existsSync(path.join(String(userDataDir), path.sep, 'schemas', `${k8sVersion}.json`));
-    },
+  const doesSchemaExistCb = useCallback(
```
topliceanurazvan, line 272: Can we remove this if is not used anymore?

topliceanurazvan (changes requested):

src/components/organisms_new/SettingsPane/Settings/Settings.tsx
mortada-codes: resolved

devcatalin (approved):

topliceanurazvan (approved):

Issue #3016, opened 2023-01-19
Show link to docs when any "unauthorized" notification errors appear


devcatalin:
If we find the word "unauthorized" in a notification error, let's add a message at the top saying:

```
We're sorry, it looks like you're not authorized to connect to this cluster. 
Please take a look at our 
[troubleshooting guide for cluster connections in our documentation](https://kubeshop.github.io/monokle/cluster-issues/) 
for steps on how to resolve this issue.

Error:
<replace with the actual error>
```

abhinandanwadwa:
Hi @devcatalin, I've tried to fix this issue in the PR #3017

PR #3017, opened 2023-01-20
fix: Show link to docs on any unauthorized notification errors.

abhinandanwadwa:
## Changes

Whenever a notification message contains the word "unauthorized", i've replaced that with the text given in the issue #3016 

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/molecules/NotificationMarkdown/NotificationMarkdown.tsx
+++ b/src/components/molecules/NotificationMarkdown/NotificationMarkdown.tsx
@@ -12,6 +12,7 @@ import store from '@redux/store';
 
 import {TelemetryButtons} from '@molecules/NotificationMarkdown/TelemetryButtons';
 
+import enhanceErrorMessage from '@utils/notification';
 import {openUrlInExternalBrowser} from '@utils/shell';
 
 import NotificationModalTitle from './NotificationModalTitle';
@@ -46,7 +47,7 @@ const NotificationMarkdown: React.FC<NotificationProps> = props => {
     Modal[type]({
       content: (
         <S.NotificationModalContent>
-          <ReactMarkdown>{message}</ReactMarkdown>
+          <ReactMarkdown>{enhanceErrorMessage(message)}</ReactMarkdown>
         </S.NotificationModalContent>
       ),
       title: (
--- a/src/components/organisms/MessageBox/MessageBox.tsx
+++ b/src/components/organisms/MessageBox/MessageBox.tsx
@@ -27,6 +27,20 @@ const MessageBox: React.FC = () => {
     [alert]
   );
 
+  const modify = (title: string) => {
+    const wordsArray = title.split(' ');
+    const unauthIndex = wordsArray.indexOf('UNAUTHORIZED');
+
+    if (unauthIndex === -1) {
+      return title;
+    }
+
+    return `We're sorry, it looks like you're not authorized to connect to this cluster. Please take a look at our troubleshooting guide for cluster connections in our documentation https://kubeshop.github.io/monokle/cluster-issues for steps on how to resolve this issue.\n\n
+      
+    Error:\n
+    ${title}`;
+  };
+
   useEffect(() => {
     if (!alert) {
       return;
@@ -36,7 +50,7 @@ const MessageBox: React.FC = () => {
       // @ts-ignore
       notification[notificationType]({
         key: alert.id,
-        message: alert.title,
+        message: modify(alert.title),
         description: <NotificationMarkdown notification={alert} type={notificationType} />,
         duration: alert.duration || 4,
       });
--- a/src/components/organisms/MessageBox/MessageBox.tsx
+++ b/src/components/organisms/MessageBox/MessageBox.tsx
@@ -29,20 +29,6 @@ const MessageBox: React.FC = () => {
     [alert]
   );
 
-  const modify = (title: string) => {
-    const wordsArray = title.split(' ');
-    const unauthIndex = wordsArray.indexOf('UNAUTHORIZED');
-
-    if (unauthIndex === -1) {
-      return title;
-    }
-
-    return `We're sorry, it looks like you're not authorized to connect to this cluster. Please take a look at our troubleshooting guide for cluster connections in our documentation https://kubeshop.github.io/monokle/cluster-issues for steps on how to resolve this issue.\n\n
-      
-    Error:\n
-    ${title}`;
-  };
-
   useEffect(() => {
     if (!alert) {
       return;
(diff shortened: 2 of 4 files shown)


src/components/organisms/MessageBox/MessageBox.tsx
```diff
@@ -27,6 +27,20 @@ const MessageBox: React.FC = () => {
     [alert]
   );
 
+  const modify = (title: string) => {
```
devcatalin, line 30: Let's create a new file `src/utils/notification.ts` and move this function there so we don't duplicate the code.
I also suggest renaming the function name to `enhanceErrorMessage` and the param to `message`.

src/components/organisms/MessageBox/MessageBox.tsx
```diff
@@ -27,6 +27,20 @@ const MessageBox: React.FC = () => {
...
+    const wordsArray = title.split(' ');
+    const unauthIndex = wordsArray.indexOf('UNAUTHORIZED');
+
+    if (unauthIndex === -1) {
+      return title;
+    }
+
+    return `We're sorry, it looks like you're not authorized to connect to this cluster. Please take a look at our troubleshooting guide for cluster connections in our documentation https://kubeshop.github.io/monokle/cluster-issues for steps on how to resolve this issue.\n\n
```
devcatalin, line 38: To make the link work properly with Markdown, we should use the markdown link notation `[text](link)`
```suggestion
    return `We're sorry, it looks like you're not authorized to connect to this cluster. Please take a look at our [troubleshooting guide for cluster connections in our documentation](https://kubeshop.github.io/monokle/cluster-issues/) for steps on how to resolve this issue.\n\n
```

devcatalin (changes requested):

src/components/organisms/MessageBox/MessageBox.tsx
```diff
@@ -27,6 +27,20 @@ const MessageBox: React.FC = () => {
     [alert]
   );
 
+  const modify = (title: string) => {
+    const wordsArray = title.split(' ');
+    const unauthIndex = wordsArray.indexOf('UNAUTHORIZED');
```
devcatalin, line 32: You don't have to split the text into words because there might be a case where we have `[Unauthorized]` in the text and then we will not catch that.

I would instead do the following:

```ts
const isUnauthorizedError = title.toLowerCase().includes('unauthorized');
```

devcatalin (changes requested):

devcatalin:
Hi @abhinandanwadwa , thank you for your contribution!
I reviewed the PR and left a few comments

src/components/organisms/NotificationsManager/Notification.tsx
```diff
@@ -60,7 +74,7 @@ const Notification: React.FC<NotificationProps> = props => {
           <S.CopyOutlined onClick={onCopyToClipboard} />
         </Tooltip>
         <S.MessageBodyContainer>
-          <S.TitleSpan>{title}</S.TitleSpan>
+          <S.TitleSpan>{modify(title)}</S.TitleSpan>
           <S.MessageSpan>
             <NotificationMarkdown notification={notification} type={notificationType} />
```
devcatalin, line 79: Now that I looked again at the code, I think the method you created will have to be called somewhere in the `NotificationMarkdown` component, on the `notification.message` property, not on the `notification.title`

src/components/organisms/MessageBox/MessageBox.tsx
```diff
@@ -36,7 +50,7 @@ const MessageBox: React.FC = () => {
       // @ts-ignore
       notification[notificationType]({
         key: alert.id,
-        message: alert.title,
+        message: modify(alert.title),
```
devcatalin, line 53: Here we probably had a problem before, I think it should've been the `alert.message` being used..

devcatalin (changes requested):

abhinandanwadwa:
Thanks for the reviews, i've started working on them!

src/components/organisms/NotificationsManager/Notification.tsx
abhinandanwadwa: so should i remove the changes from the `Notification` and `MessageBox` component?

src/components/organisms/MessageBox/MessageBox.tsx
```diff
@@ -36,7 +38,7 @@ const MessageBox: React.FC = () => {
       // @ts-ignore
       notification[notificationType]({
         key: alert.id,
-        message: alert.title,
+        message: enhanceErrorMessage(alert.title),
```
topliceanurazvan, line 41: This here shouldn't be enhanced since it's the title, just the description should be modified

src/components/organisms/MessageBox/MessageBox.tsx
```diff
@@ -36,7 +38,7 @@ const MessageBox: React.FC = () => {
...
         description: <NotificationMarkdown notification={alert} type={notificationType} />,
```
topliceanurazvan, line 42: I would put here instead of `notification={alert}` -> `notification={{...alert, message: enhanceErrorMessage(alert.message)}}`, so in this way you don't have to modifiy anything inside of the `NotificationMarkdown` component

src/components/organisms/NotificationsManager/Notification.tsx
```diff
@@ -60,7 +62,7 @@ const Notification: React.FC<NotificationProps> = props => {
           <S.CopyOutlined onClick={onCopyToClipboard} />
         </Tooltip>
         <S.MessageBodyContainer>
-          <S.TitleSpan>{title}</S.TitleSpan>
+          <S.TitleSpan>{enhanceErrorMessage(title)}</S.TitleSpan>
```
topliceanurazvan, line 65: This here must be left only as `{title}`

topliceanurazvan (changes requested):
Hey @abhinandanwadwa, I've left a couple of comments on your code

src/components/molecules/NotificationMarkdown/NotificationMarkdown.tsx
```diff
@@ -46,7 +47,7 @@ const NotificationMarkdown: React.FC<NotificationProps> = props => {
     Modal[type]({
       content: (
         <S.NotificationModalContent>
-          <ReactMarkdown>{message}</ReactMarkdown>
+          <ReactMarkdown>{enhanceErrorMessage(message)}</ReactMarkdown>
```
topliceanurazvan, line 50: With the change on `MessageBox`, this can be just `{message}`

src/components/organisms/NotificationsManager/Notification.tsx
```diff
@@ -60,7 +62,7 @@ const Notification: React.FC<NotificationProps> = props => {
...
           <S.MessageSpan>
             <NotificationMarkdown notification={notification} type={notificationType} />
           </S.MessageSpan>
```
topliceanurazvan, line 68: Same as above here in `Notification`, this should be changed `notification={alert} -> notification={{...alert, message: enhanceErrorMessage(alert.message)}}`

PR #3051, opened 2023-01-30
fix: add walkthrough cards deps

mortada-codes:
This PR...

## Changes

- added walkthroughs cards to learn page

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms_new/StartPage/StartPage.tsx
```diff
@@ -1,10 +1,13 @@
 import {useEffect, useState} from 'react';
 
+import {WALK_THROUGH_STEPS} from '@constants/walkthrough';
```
topliceanurazvan, line 3: We should add this inside `shared` folder as most probably `@constants` will be removed in the near future

src/index.css
```diff
@@ -450,3 +450,11 @@ div.monokleEditorErrorRefGlyphClass.monokleEditorUnsatisfiedRefGlyphClass::after
 .terminal-options-menu .ant-dropdown-menu-item:hover {
   background-color: transparent;
 }
+
+.walkthrough-modal ul li::marker {
+  color: #1d1d1d;
+}
+
+.walkthrough-modal .ant-modal-content .ant-modal-body {
```
topliceanurazvan, line 458: Is this something that can be fixed inside the core component also? 🤔

topliceanurazvan (changes requested):

src/index.css
mortada-codes: We cannot fix it from there because the Monokle desktop css is overriding the modal background color

topliceanurazvan (approved):

PR #3098, opened 2023-02-06
feat: sarif viewer

topliceanurazvan:
## Changes

- Initial validation of resources
- Listener when k8s-schema changes/plugin is turned off/on/rule toggle
- Add SARIF editor when selecting a problem

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/PageHeader/K8sVersionSelection.tsx
```diff
@@ -47,7 +50,7 @@ export const K8sVersionSelection = () => {
       <S.MenuBottom>
         <S.WarningText>
           Selecting another K8s Schema can bring <Typography.Text type="danger">new validation errors.</Typography.Text>{' '}
-          Check them out in from the left menu.
+          Check them out in the left menu.
```
devcatalin, line 53:
```suggestion
          Check them out in the Validation panel from the left menu.
```

src/components/organisms_new/ExplorerPane/FilePane/FilePane.tsx
```diff
@@ -251,7 +251,7 @@ const FilePane: React.FC = () => {
             <span id="file-explorer-count">
               <b>{filesOnly.length || 0} files</b>
             </span>{' '}
-            in <span id="file-explorer-project-name">{fileMap[ROOT_FILE_ENTRY].filePath}</span>
+            in <span id="file-explorer-project-name">{fileMap[ROOT_FILE_ENTRY]?.filePath}</span>
```
devcatalin, line 254:
```suggestion
            in <span id="file-explorer-project-name">{rootEntry.filePath}</span>
```

src/redux/validation/validation.listeners.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+import {setConfigK8sSchemaVersion, toggleOPARules, toggleValidation} from './validation.slice';
+import {loadValidation, validateResources} from './validation.thunks';
+
+const loadListener: AppListenerFn = listen => {
+  listen({
+    matcher: isAnyOf(
+      loadClusterResources.fulfilled,
+      reloadClusterResources.fulfilled,
+      setRootFolder.fulfilled,
```
devcatalin, line 29: do we have to load the validator for `loadClusterResources` and `reloadClusterResources` ?
maybe for the `loadClusterResources` it makes sense if it's triggered by the quick cluster loading?

src/redux/validation/validation.listeners.tsx
```diff
@@ -0,0 +1,133 @@
...
+      const loading = dispatch(loadValidation());
+      signal.addEventListener('abort', () => loading.abort());
+      await loading;
+    },
+  });
+};
+
+const validateListener: AppListenerFn = listen => {
+  listen({
+    matcher: isAnyOf(loadValidation.fulfilled),
```
devcatalin, line 46: I'm guessing here is where we'll have to add the list of all actions where we actually want to run the validation, and not necessarily rely only on loadValidation.fulfilled, but I can tackle this one in a separate PR

src/redux/validation/validation.listeners.tsx
```diff
@@ -0,0 +1,133 @@
...
+      const state = getState();
+
+      const activeProject = activeProjectSelector(state);
+      const clusterResourceMap = clusterResourceMapSelector(state);
+      const currentContext = kubeConfigContextSelector(state);
+      const localSchemaVersion = currentConfigSelector(getState()).k8sVersion;
+      const userDataDir = String(state.config.userDataDir);
+
+      const nodeResources = Object.values(clusterResourceMap).filter(
+        resource => resource.object.apiVersion === 'v1' && resource.kind === 'Node'
```
devcatalin, line 82:
```suggestion
        resource => resource.apiVersion === 'v1' && resource.kind === 'Node'
```

src/redux/validation/validation.listeners.tsx
```diff
@@ -0,0 +1,133 @@
...
+      );
+
+      if (isEmpty(nodeResources)) return;
+
+      const clusterSchemaVersions = [
+        ...new Set(
+          nodeResources.map(resource =>
+            resource.object?.status?.nodeInfo?.kubeletVersion?.split('+')[0].substring(1).trim()
+          )
+        ),
```
devcatalin, line 92: I would suggest doing the following to avoid error like "cannot call .split() on undefined".
I'm also not sure if the kubeletVersion is in the same object path on all k8s schema versions, but we could leave this solution for now
```suggestion
        ...new Set(
          nodeResources.map(resource => {
            const kubeletVersion = esource.object?.status?.nodeInfo?.kubeletVersion;
            if (typeof kubeletVersion !== 'string') {
              return undefined;
            }
            return kubeletVersion.split('+')[0].substring(1).trim()
          }).filter(isDefined)
        ),
```

devcatalin (changes requested):

devcatalin (approved):

PR #3119, opened 2023-02-08
fix: update bg color across the app

mortada-codes:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms_new/PaneManager/PaneManagerLeftMenu.tsx
```diff
@@ -22,34 +25,42 @@ const NewPaneManagerLeftMenu: React.FC = () => {
...
+          dispatch(setLeftBottomMenuSelection(activityName));
+        }}
+      />
+    </Container>
   );
 };
 
 export default NewPaneManagerLeftMenu;
+
+const Container = styled.div`
```
topliceanurazvan, line 62: Can't we just use the `style` prop for the `ActivityBar` directly in here?

topliceanurazvan (changes requested):

src/components/organisms_new/PaneManager/PaneManagerLeftMenu.tsx
mortada-codes: We could use it if we added a class name or made it accept the style prop

topliceanurazvan (dismissed):

topliceanurazvan (approved):

PR #3121, opened 2023-02-08
fix: long running loading 

mortada-codes:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms_new/CompareSyncPane/CompareModalSelecting/CompareModalSelecting.tsx
```diff
@@ -68,11 +69,15 @@ const CompareModalSelecting: React.FC = () => {
...
             {right.loading ? (
-              <Skeleton active />
+              <div style={{paddingLeft: hasSideSelected ? 6 : 0, height: '100%', overflow: 'hidden'}}>
+                <Loading />
+              </div>
             ) : right.error ? (
               <ErrorFigure onRetry={() => handleRetry('right')} />
             ) : (
-              <ResourceList data={right} />
+              <div style={{paddingLeft: 6}}>
```
topliceanurazvan, line 78: Can we add this padding directly in the `ResourceList` component? 🤔

topliceanurazvan (changes requested):

src/components/organisms_new/CompareSyncPane/CompareModalSelecting/CompareModalSelecting.tsx
mortada-codes: We cannot because it depends on the component's position. On the left hand, we don't need the padding, but on the right hand, we need to add it because of the scrollbar width

topliceanurazvan (approved):

PR #3144, opened 2023-02-10
fix:   maximum hook update

mortada-codes:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


devcatalin:
I've made some changes thinking that this might fix the FormEditor but it looks like the Form is still infintely re-rendering for some reason. We'll have to test it more.

mortada-codes:
> I've made some changes thinking that this might fix the FormEditor but it looks like the Form is still infintely re-rendering for some reason. We'll have to test it more.

working on it

src/components/molecules/FormEditor/FormEditor.tsx
```diff
@@ -92,30 +82,7 @@ const FormEditor: React.FC<IProps> = props => {
...
-      const fileContent = fs.readFileSync(absoluteFilePath, 'utf8');
-      const content = mergeManifests(fileContent, formString);
-      return content.trim() !== fileContent.trim();
-    }
-
-    return false;
-  }, [fileMap, formData, selectedFilePath, selectedResource?.text]);
-
-  useWhatChanged([fileMap, selectedFilePath, selectedResource, formData, changed]);
+  useWhatChanged([fileMap, selectedFilePath, selectedResource, formData]);
```
devcatalin, line 85: Let's remove this now since it was only used for debugging!

devcatalin (changes requested):

src/redux/thunks/saveResourceChanges.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import {setAutosavingError, setAutosavingStatus} from '@redux/reducers/main';
+import {selectedFilePathSelector} from '@redux/selectors';
+import {getAbsoluteFilePath} from '@redux/services/fileEntry';
+import {mergeManifests} from '@redux/services/manifest-utils';
+
+import {AppDispatch} from '@shared/models/appDispatch';
+import {RootState} from '@shared/models/rootState';
+
+export const saveResourceChanges = createAsyncThunk<void, string, {dispatch: AppDispatch; state: RootState}>(
```
devcatalin, line 14: Let's rename this to `saveFormEditorResource` so we don't confuse it with something else

src/redux/thunks/readResourceFile.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import fs from 'fs/promises';
+import log from 'loglevel';
+
+import {getAbsoluteFilePath} from '@redux/services/fileEntry';
+
+import {AppDispatch} from '@shared/models/appDispatch';
+import {RootState} from '@shared/models/rootState';
+
+export const readResourceFile = createAsyncThunk<string | undefined, string, {dispatch: AppDispatch; state: RootState}>(
```
devcatalin, line 11: We can rename this to `readFileThunk` since it only reads a file

devcatalin (changes requested):

devcatalin (approved):

PR #3171, opened 2023-02-12
feat: add BackToProjects button to quick cluster preview

erdkse:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan (changes requested):
Looks good, the only thing here is that if we have an active project and we go for `Quick cluster preview`, we should still show the same header from the Editor UI, not the back to projects page. Also we should still see all the left menu options ( they should be hidden only if we don't have an active project currently )

topliceanurazvan (approved):

PR #3247, opened 2023-02-20
fix: handling InputNumber onStep

mortada-codes:
This PR...

## Changes

- Form onChange wasn't handling the InputNumber onStep event

- onValuesChange get fired before the validation so we have to manually validate the fontsize
## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan (changes requested):
Looks good, but for some reason when I click on the input to change the font size, the popup closes. Can we fix that also? 🤔

topliceanurazvan (approved):

PR #3372, opened 2023-03-09
feat: handle Image pull errors

erdkse:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


src/components/organisms/Dashboard/Dashboard.tsx
```diff
@@ -57,21 +57,21 @@ import {Tableview} from './Tableview/Tableview';
 const Dashboard: React.FC = () => {
   const activeMenu = useAppSelector(state => state.dashboard.ui.activeMenu);
   const menuList = useAppSelector(state => state.dashboard.ui.menuList);
-  const clusterResourceContentMapRef = useResourceContentMapRef('cluster');
+  const clusterResourceContent = useAppSelector(state => state.main.resourceContentMapByStorage['cluster']);
```
erdkse, line 60: @devcatalin Can you look here? `useResourceContentMapRef ` was not updating the resources, that is why I canceled it

devcatalin: Makes sense, when I did the refactor I was not sure when this should be recomputed, so I thought it should be o
only when the `activeMenu` changes.
Maybe you should do the same for clusterResourceMetaMap ?

also please use the selectors:
```ts
const clusterResourceMetaMap = useResourceMetaMap('cluster');
const clusterResourceContentMap = useResourceContentMap('cluster');
```

devcatalin (changes requested):

mortada-codes (dismissed):

PR #3378, opened 2023-03-12
Fixed clicking outside resource label not selecting the item

KarthikRaju391:
This PR...

## Fixes

- Clicking on the resource label not selecting the item

## How to test it

- Click anywhere on the resource label

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan (changes requested):
Looks good, the only think that we still need to do is not highlight the element when clicking on the three dots for showing the menu ( this should happen for all section blueprints that contains `ContextMenu`

KarthikRaju391:
@topliceanurazvan, so, when we click on the three dots, the item shouldn't be highlighted right?

topliceanurazvan:
> @topliceanurazvan, so, when we click on the three dots, the item shouldn't be highlighted right?

Yes

KarthikRaju391:
> @topliceanurazvan, so, when we click on the three dots, the item shouldn't be highlighted right?

@topliceanurazvan, I am not able to find a way to make it that way

topliceanurazvan:
> > @topliceanurazvan, so, when we click on the three dots, the item shouldn't be highlighted right?
> 
> @topliceanurazvan, I am not able to find a way to make it that way

Hey @KarthikRaju391 , a solution would be to stop the event propagation on the click of the three dots

KarthikRaju391:
@topliceanurazvan, I think I have fixed the even propogation problem

topliceanurazvan (approved):
Looks good to me! Great work 🙌🏻

PR #3465, opened 2023-03-22
fix: added fail events and fixed missing metadata

olensmar:
This PR adds additional telemetry for failed previews - so we can get a better understanding of why they fail and perhaps handle those cases better


mortada-codes (dismissed):

src/redux/thunks/runPreviewConfiguration.ts
```diff
@@ -122,26 +122,29 @@ export const runPreviewConfiguration = createAsyncThunk<
   const result = await runCommandInMainThread(commandOptions);
 
   if (result.error) {
+    trackEvent('preview/helm_config/fail', {reason: result.error});
```
olensmar, line 125: @devcatalin @WitoDelnat my only concern here is that the error could contain sensitive info that we shouldn't be collecting... but we would like to know why a preview fails so we can perhaps but some functionality in place to help.. wdyt?

olensmar: i don't think this gets logged to Sentry - as it's not an uncaught exception. And the business might care if the reason is something we can work with - i.e. missing dependencies (which we could check for automatically), missing values, etc..

WitoDelnat: I think you can use Sentry client to manually forward errors as well. Benefit is that Sentry includes other details such as Stack trace, browser and more. While I'm slightly in favour of decoupling the two, I have no strong opinion so I'd definitely also go ahead with what we got here ✅

PR #3572, opened 2023-03-31
feat: add ipc framework

WitoDelnat:
This PR adds a framework to deal with IPC in a robuust way.

### DIFF
--- a/electron/app/services/cluster/ipc.ts
+++ b/electron/app/services/cluster/ipc.ts
@@ -1,0 +1,4 @@
+import {handleIpc} from '../../utils/ipc';
+import {ping} from './handlers/ping';
+
+handleIpc('cluster:ping', ping);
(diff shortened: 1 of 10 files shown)


src/redux/cluster/service/ping.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * ```
+ * try {
+ *   const { ok } = await ping({ context: "minikube" })
+ * } catch (err) {
+ *   console.log(err);
+ * }
+ * ```
+ */
+export const ping = invokeIpc<PingParams, PingResult>('cluster:ping');
```
WitoDelnat, line 16: Step 1. You define your typed invocation helper like this.

electron/app/services/cluster/ipc.ts
```diff
@@ -0,0 +1,8 @@
+import {PingParams} from '@shared/ipc';
+
+import {handleIpc} from '../../utils/ipc';
+import {ping} from './handlers/ping';
+
+handleIpc('cluster:ping', (params: PingParams) => {
```
WitoDelnat, line 6: Step 2. You define your handler helper like this.

Errors thrown from the handler will be properly forwarded to the renderer and rethrown on that side.

devcatalin (approved):
LGTM

PR #3623, opened 2023-04-05
chore: remove bulk cluster connection notifications on access check

erdkse:
This PR throws a custom error on Cluster Connection error on the namespace access check and it doesn't throw notifications for every namespace. So hopefully it will not bother the users.

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/shared/utils/kubeclient.ts
+++ b/src/shared/utils/kubeclient.ts
@@ -125,7 +125,12 @@ export const getKubeAccess = async (namespace: string, currentContext: string): 
   if (hasErrors) {
     const errors = result.stderr;
     log.error(`get cluster access errors ${errors}`);
-    throw new Error(`Couldn't get cluster access for namespaces. Errors: ${JSON.stringify(errors)}`);
+    // Shall we create CustomErrors to handle different error types?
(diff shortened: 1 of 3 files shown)


src/shared/utils/kubeclient.ts
```diff
@@ -125,7 +125,12 @@ export const getKubeAccess = async (namespace: string, currentContext: string):
   if (hasErrors) {
     const errors = result.stderr;
     log.error(`get cluster access errors ${errors}`);
-    throw new Error(`Couldn't get cluster access for namespaces. Errors: ${JSON.stringify(errors)}`);
+    // Shall we create CustomErrors to handle different error types?
```
erdkse, line 128: Shall we handle situations like this with Custom Errors? @devcatalin @topliceanurazvan @mortada-codes

devcatalin: this looks okay to me

devcatalin (approved):

PR #3670, opened 2023-04-09
fix: remove Events from left menu of the Cluster Dashboard

erdkse:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
if we remove the events from here - where would a click on an event in the overview go? how can I see details of an event?

PR #3734, opened 2023-04-14
fix: validation resources count

mortada-codes:
This PR...

## Changes

-

## Fixes

- applying resource filter on errors count and warnings count

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan (changes requested):
On initial load, the total number of warnings is 38, but the count from the navigator shows only 6 ( without applying any filtering to resources )

image

src/redux/selectors/resourceSelectors.ts
```diff
@@ -246,3 +247,17 @@ export const resourceNavigatorSelector = createSelector(
...
+    Object.values(resources)
+      .filter(
+        r =>
+          r.kind &&
+          isResourcePassingFilter(r, resourceFilter) &&
+          !isKustomizationResource(r) &&
+          !isKustomizationPatch(r)
+      )
+      .map(r => r.id)
+);
```
devcatalin, line 263: Now that Ole's PR is merged we have a new `navigatorResourcesSelector` which could be used here because it merges the active storage together with transient resources and applies the filtering:

```suggestion
export const filteredResourcesIdsSelector = createSelector(
  navigatorResourcesSelector,
  (resources) => Object.values(resources).map(r => r.id)
);
```

src/redux/validation/validation.selectors.ts
```diff
@@ -67,6 +67,9 @@ export const problemsByResourcesSelector = createDeepEqualSelector(
 export const errorsByResourcesSelector = (state: ValidationState) => problemsByResourcesSelector(state, 'error');
 export const warningsByResourcesSelector = (state: ValidationState) => problemsByResourcesSelector(state, 'warning');
 
+export const errorsResourcesIdsSelector = (state: ValidationState) => Object.keys(errorsByResourcesSelector(state));
+export const warningsResourcesIdsSelector = (state: ValidationState) => Object.keys(warningsByResourcesSelector(state));
```
devcatalin, line 71: Maybe it helps to have `createSelector` here as well for the memoization, or we could even use `createDeepSelector` 🤔 
```suggestion
export const errorsResourceIdsSelector = createSelector(errorsByResourcesSelector, errors => Object.keys(errors));
export const warningsResourcesIdsSelector = createSelector(warningsByResourcesSelector, warnings => Object.keys(warnings));

```

devcatalin (approved):

topliceanurazvan (approved):

PR #3948, opened 2023-05-15
refactor: is hovered renderers + change preview to dry-run

topliceanurazvan:
## Changes

- Show the context menu dots only on hover
- Change `Preview` label to `Dry-run`

## Screenshots

image

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/PageHeader/PreviewControl/PreviewLabel.tsx
+++ b/src/components/organisms/PageHeader/PreviewControl/PreviewLabel.tsx
@@ -22,7 +22,7 @@ export function PreviewLabel() {
       <LabelBox>
         {PREVIEW_ICON_MAP[preview.type]}
         <LabelContent>
-          Previewing <span style={{fontWeight: 700}}>{previewDisplayContent.name}</span>
+          Dry-running <span style={{fontWeight: 700}}>{previewDisplayContent.name}</span>
         </LabelContent>
       </LabelBox>
     </Tooltip>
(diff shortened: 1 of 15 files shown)


olensmar:
TBH it looks a little weird to me - but that's probably just because I'm so used to "Preview" :D

olensmar:
@CortNick would this change require any urgent updates to website (or any other) content?

src/components/organisms/PageHeader/PreviewControl/PreviewLabel.tsx
```diff
@@ -22,7 +22,7 @@ export function PreviewLabel() {
       <LabelBox>
         {PREVIEW_ICON_MAP[preview.type]}
         <LabelContent>
-          Previewing <span style={{fontWeight: 700}}>{previewDisplayContent.name}</span>
+          Dry-running <span style={{fontWeight: 700}}>{previewDisplayContent.name}</span>
```
olensmar, line 25: "Performing Dry-run" ? or is that too long?

olensmar: oh - then "Viewing Kustomize dry run" is probably better - and "Viewing Helm-chart dry run" ? or "Showing.. " ? @CortNick wdyt? 

Also - grammatically it seems to be more correct to not hyphenise - i.e. "dry run" - but that looks even weirder!? And should we capitalize? i.e. "dry-run" or "Dry-run"?

olensmar (approved):

PR #3952, opened 2023-05-16
fix: catch crd dir exists error

mortada-codes:
This PR...

## Changes

- added tr/catch cluse to handle EEXIST: file already exists. 

## Fixes

-  fixing a race condition while saving many crds at once 

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/utils/crds.ts
+++ b/src/utils/crds.ts
@@ -12,7 +12,7 @@ export async function saveCRD(crdsDir: string, crdContent: string) {
       await createFolder(crdsDir);
     }
   } catch (error) {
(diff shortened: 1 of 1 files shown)


mortada-codes:
it is better to check if the user has access to the `crdDir` and halt the saving operation if the user has no access privilege but `doesPathExist` doesn't throw an error instead it returns false.

olensmar (approved):

PR #3985, opened 2023-05-24
fix: add modal confirms when transferring in compare

topliceanurazvan:
## Fixes

- Show a confirm when trying to transfer a resource in the compare & sync
- Show back the extract buttons inside the inspection ( diff )

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/components/organisms/CompareSyncPane/ComparisonList/ComparisonList.tsx
+++ b/src/components/organisms/CompareSyncPane/ComparisonList/ComparisonList.tsx
@@ -1,6 +1,6 @@
 import {useCallback} from 'react';
 
-import {Button, Checkbox, Col} from 'antd';
+import {Button, Checkbox, Col, Modal} from 'antd';
 
 import {
   comparisonInspecting,
@@ -10,6 +10,8 @@ import {
   transferResource,
 } from '@redux/compare';
 import {useAppDispatch, useAppSelector} from '@redux/hooks';
+
+import {useRefSelector} from '@utils/hooks';
 
 import {CompareSide, ComparisonItemProps, HeaderItemProps} from '@shared/models/compare';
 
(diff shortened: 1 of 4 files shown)


src/components/organisms/CompareSyncPane/ComparisonList/ComparisonList.tsx
```diff
@@ -68,13 +70,23 @@ function ComparisonItem({
...
   const handleTransfer = useCallback(
     (side: CompareSide) => {
       const direction = side === 'left' ? 'left-to-right' : 'right-to-left';
-      dispatch(transferResource({ids: [id], direction}));
+      const transferTo = side === 'left' ? rightTypeRef.current : leftTypeRef.current;
+
+      Modal.confirm({
+        title: `Are you sure you want to ${transferTo === 'cluster' ? 'deploy to cluster' : 'extract to local'}?`,
+        onOk() {
+          dispatch(transferResource({ids: [id], direction}));
```
olensmar, line 84: telemetry here? include the transferTo and number of resources if possible?

src/components/organisms/CompareSyncPane/TransferButton/TransferButton.tsx
```diff
@@ -43,8 +43,14 @@ const TransferButton: React.FC<Props> = ({side}) => {
   });
 
   const handleTransfer = useCallback(() => {
-    dispatch(transferResource({ids, direction}));
-  }, [direction, dispatch, ids]);
+    Modal.confirm({
+      title: `Are you sure you want to ${buttonLabel.toLowerCase()}?`,
+      onOk() {
+        dispatch(transferResource({ids, direction}));
```
olensmar, line 49: as above..

topliceanurazvan:
We already have telemetry in place for this, it's inside the compare slice, under `transferResource.fulfilled`

image

olensmar:
> 

ok thanks - I think we should strive to do telemetry as close to the user action as possible (and not in slices) so we can see where users are performing actions - but we don't have to change this now - just keep in mind for future telemetry!

PR #3992, opened 2023-05-25
feat: add graph view in cluster dashboard

topliceanurazvan:
## Changes

- Add graph view tab in cluster dashboard drawer

## Screenshots

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
thanks @topliceanurazvan - a couple of suggestions/issues:
- the pan/zoom seems to reset itself every x seconds - guessing because of some background update?
- could we implement that double-clicking a component selects that component in dashboard (i.e. the corresponding kind and resource)
- would be great if it was possible to resize (drag) the panel so the graph view can be made larger - it seems the right panel has a fixed size now
- could we default to "Max 1 link deep" ?
- with a small window I get the below without the possibility to scroll/move the info panel into view.. not sure what the behaviour should be here!?

olensmar (approved):

Issue #3695, opened 2023-04-10
excessive space between accordion sections


olensmar:
see below - there is some additional space at the bottom of each section before the next header

<img width="553" alt="image" src="https://user-images.githubusercontent.com/1917063/230892396-0b36c8bf-5c6e-4ad9-a46b-cad7c36e59dd.png">

itsajay1029:
Can you assign this to me @olensmar ?

olensmar:
sure - thanks!!

itsajay1029:
Where will I find the deployment Patches section @olensmar ? Can you ellaborate a little on the issue please ?

olensmar:
sure @itsajay1029: there is some empty space above the headers of each accordion section (except the first one) - as outlined in the below image - the actual content of the section above (in this case "Deployment Patches") doesn't matter - you just need to find the styling in the accordion that needs to be adjusted accordingly.

<img width="577" alt="image" src="https://user-images.githubusercontent.com/1917063/231080348-901a14ed-f928-46a5-9004-62d5fe7a6743.png">

ayushgml:
@olensmar I want to work on this please assign me this issue. Thank you!

ayushgml:
Hey @olensmar I worked on it and this is how the gap is(Like the first section):

<img width="354" alt="Screenshot 2023-05-29 at 4 01 37 PM" src="https://github.com/kubeshop/monokle/assets/72748253/81e3d9b6-c3b3-4ab3-b18e-ed88d3845697">

What I did was: I used display flex and justify-content as space-between(for Accordion Panel). Then I went to increase the Accordion height by 12px.

topliceanurazvan:
Hello @ayushgml, can you create a PR for this fix so we can give it a try too? Also, the spacing should be a little bit more, with some padding between the actually opened accordion and the next one.

olensmar:
@julioramosest can you provide an exact number for the spacing? (in pixels!?)

ayushgml:
Yeah @topliceanurazvan I will create a PR so you can try it too!

julioramosest:
[...]

PR #4015, opened 2023-05-29
Excessive space accordion

ayushgml:
This PR addresses the issue:

 - #3695 

## Changes

- I added the following in ```ant-collapse-item-active``` (Panel) : 
```
display: flex;
flex-direction: column;
justify-content: space-between;
```

## How to test it

- Run the Electron app and go to the ExplorerPane

## Screenshots

Screenshot 2023-05-29 at 4 01 37 PM

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


ayushgml:
Hey @topliceanurazvan! I made the necessary changes. Please suggest any changes if you encounter any issues.

topliceanurazvan (changes requested):
Seems like we are still missing some padding from the bottom of the pane:

image

Expected:

image

The rest of the work looks good! 🙌🏻

ayushgml:
Okay @topliceanurazvan I will fix this padding issue!

ayushgml:
@topliceanurazvan I made the change! The padding issue was happening only when FilesPane was selected and padding was there when other panes were selected. This was because FilesPane has a slightly different structure than the rest of them. Rest of them have lists which are made using the styled component ListContainer. FilesPane didn't have it so i removed the extra height only when the FilesPane is selected.

This issue issue might be due to the height miscalculation in the ```AccordionPanel```:
```const height = usePaneHeight() - PANEL_HEADER_HEIGHT - 130 - (props.panelKey === 'files' ? 35 : 25);```

Though I fixed the issue!🙌🏼 You can test it out from your end.

topliceanurazvan (approved):
Great work! 👍🏻

PR #4020, opened 2023-05-30
refactor: new resource creation

topliceanurazvan:
## Changes

- Replace the collapse/expand icon for both the navigator and file tree
- Remove new folder from file tree header
- Change plus button for creating a new resource with label `New`
- Change new resource dropdown items label
- Add new resource cards when the navigator is empty
- Remove `highlightedItems`

## Screenshots

image

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


topliceanurazvan:
@julioramosest @olensmar @devcatalin Let me know if we need to change the labels from the new resource dropdown items

julioramosest:
- Not sure about the radius and the font weight of the new CTA. Radius needs to be 3px, font is semibold, 12px
- the expand icon in the resources column. Needs to be link-blue by default

as for the labels: Thomas suggested "New from model" to be "Create manually" or sort of...

olensmar:
@DzoQiEuoi @CortNick  do you have some suggestions on how to improve these labels? I was thinking something in line of

New
- From scratch / schema
- From advanced template
- Using GPT query

devcatalin:
I like the options mentioned by Ole, although I would go for "Using AI prompt" as the third one.

The "New from model" option is the most confusing one at the moment, at least for me.
I like the "New from scratch" better, or we could say "Using general form" 🤔

DzoQiEuoi:
I was thinking:

- Create manually
- Use a template
- Generate via AI

olensmar:
ok since everyone has different suggestions i'll make the call:
- Create manually
- Use Advanced Template
- AI Assisted

mortada-codes (approved):

PR #4023, opened 2023-05-30
feat: confirmation modal for unsaved cluster resource changes

devcatalin:
This PR...

## Changes

- added warning modal when users tries to close the resource drawer and they have an unsaved change

## How to test it

- make changes to a resource in Cluster Insights
- click anywhere outside the drawer
- the new popups should show up to ask for handling the unsaved changes

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/src/editor/editor.cluster.tsx
+++ b/src/editor/editor.cluster.tsx
@@ -72,14 +72,11 @@ const WarnUnsavedChangesModal = (props?: WarnUnsavedChangesModalProps) => {
             Deploy changes
           </Button>
           <Button onClick={onDiscard}>Discard changes</Button>
-          <Button onClick={onClose}>Cancel</Button>
         </div>
       }
     >
       <div>
-        <p>Pick what you would like to do with your changes.</p>
-        <p>If you deploy them, they will be applied to the cluster.</p>
-        <p>If you discard them, they will be lost.</p>
+        <p>You have unsaved changes.</p>
       </div>
     </Modal>
   );


src/editor/editor.cluster.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const WarnUnsavedChangesModal = (props?: WarnUnsavedChangesModalProps) => {
+  const {open, onClose} = props || {};
+  return (
+    <Modal
+      open={open}
+      centered
+      footer={
+        <div>
+          <Button onClick={onClose}>Cancel</Button>
```
olensmar, line 22: perhaps the reverse order!? Deploy, Discard, Cancel ?

src/editor/editor.cluster.tsx
```diff
@@ -0,0 +1,53 @@
...
+          <Button>Discard changes</Button>
+          <Button>Deploy changes</Button>
+        </div>
+      }
+    >
+      <div>
+        <p>Pick what you would like to do with your changes.</p>
+        <p>If you deploy them, they will be applied to the cluster.</p>
```
olensmar, line 30: append " and replace any existing object with the same name" ? or is that too obvious?

src/editor/editor.cluster.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const useWarnUnsavedChanges = (): [() => boolean, () => JSX.Element] => {
+  const [isOpen, setIsOpen] = useState(false);
+  const ModalComponent = useModalComponent({open: isOpen, onClose: () => setIsOpen(false)});
+
+  const warn = useCallback(() => {
+    if (!didEditorContentChange() || isOpen) {
+      return false;
+    }
+    alert('You have unsaved changes in your code editor.');
```
olensmar, line 47: should we use an antd alert instead?

devcatalin: Using `alert` is the only way to block all clicks outside of the drawer from going through so we have to use it unfortunately

mortada-codes (approved):

PR #4027, opened 2023-06-01
fix: update @monokle/validation 0.17.4

mortada-codes:
## Changes

- Update to latest validation/components core packages
- Show `Practices` and `Pod Security Standards` plugins
- Custom table for plugins supporting to configure them

## Fixes

- Multiple styling fixes

## Screenshots

image

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
LGTM but we need @WitoDelnat and @devcatalin to weigh in..

WitoDelnat:
Are you running in a web worker? We had to make [a quickfix](https://github.com/kubeshop/monokle-saas/pull/1370/files) because we broke plugin loading there.

topliceanurazvan (approved):

olensmar:
I'm not exactly sure what to expect here - but it seems to me that the new default validators don't show up with corresponding configuration panels - guessing that needs to be addressed before this can be merged? @WitoDelnat @mortada-codes @devcatalin @topliceanurazvan

WitoDelnat:
That should definitely be addressed, in Monokle Cloud we do this as follows:

```typescript
function ValidationPane() {
  // ...
const ConfigPane = useMemo(() => {
  return pluginName
    ? VALIDATION_CONFIGURATION_COMPONENTS[pluginName] ?? ValidationCustom
    : undefined;
}, [pluginName]);

  //...
}

export const VALIDATION_CONFIGURATION_COMPONENTS: Record<
  string,
  ((props: PluginMetadataWithConfig) => JSX.Element) | false | undefined
> = {
  'open-policy-agent': props => <ValidationOpenPolicyAgent {...props} />,
  'kubernetes-schema': props => <ValidationKubernetesSchema {...props} />,
  'resource-links': false,
  'yaml-syntax': false,
};

```

topliceanurazvan:
@olensmar I had a huddle with Wito and it seems like this needs some more work, like adding the possibility to add custom plugins and core refactoring for filters and icons... so we might have to postpone this for the next release...

olensmar:
thanks @topliceanurazvan for looking into it - that's almost what I was expecting - no problem! Should we close this for now? or change it to Draft?

src/shared/constants/validation.ts
```diff
@@ -1 +1,8 @@
-export const CORE_PLUGINS = ['open-policy-agent', 'kubernetes-schema', 'yaml-syntax', 'resource-links'];
+export const CORE_PLUGINS = [
```
olensmar, line 1: do we really have to hard-code this list? I'm sure these will change over time in core - and we might forget keeping it in sync..

olensmar (commented):
minor comments on code.. also wondering if we can rename the "Practices" plugin to "Best Practices"

WitoDelnat:
> "Practices" plugin to "Best Practices"

It started as Best Practices, but whether setting something like cpu limits is highly contested as a best practice which is why I kept it more general and a "choose and pick from common practices" to your preferred style.

src/shared/models/validationPlugins.ts
```diff
@@ -1,14 +1,9 @@
 import {IconNames} from '@monokle/components';
 
-type ValidationIntegrationId =
-  | 'kubernetes-schema'
-  | 'open-policy-agent'
-  | 'resource-links'
-  | 'yaml-syntax'
-  | 'crd-schema';
+type ValidationPluginId = 'kubernetes-schema' | 'open-policy-agent' | 'resource-links' | 'yaml-syntax' | 'crd-schema';
```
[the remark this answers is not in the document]
topliceanurazvan: This one actually is not used anymore for the validation plugins... We'll have to keep it just for the our custom CRD in order to have a type for it, I'll adjust it accordingly and change the name of the type

src/shared/constants/validation.ts
topliceanurazvan: One solution would be to have this list directly in the validation and use it from there. Right now we need this kind of list to load the initial validation with the core plugins...

olensmar:
thanks for the clarification @WitoDelnat - would "Common Practices" be ok? - just "Practices" feels like its missing something..

WitoDelnat:
That makes a lot of sense. I agree practices on it's own feels like its missing something.

Do we only want to update the display name or also the plugin identifyer (used for config file)? There I feel shorter is better.

```
rules:
  practices/no-latest-image
  common-practices/no-latest-image
```

PR #4093, opened 2023-06-20
improve telemetry

devcatalin:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


mortada-codes (dismissed):

src/components/organisms/CreateProjectModal/CreateProjectModal.tsx
```diff
@@ -49,6 +49,7 @@ const CreateProjectModal: React.FC = () => {
         dispatch(setCreateProject({name, rootFolder}));
         onCloseModalHandler();
       } else {
+        trackEvent('app_start/create_project', {from: 'folder'});
```
olensmar, line 52: shouldn't this be `template`? and perhaps we already do that further down-stream?

src/shared/models/telemetry.ts
```diff
@@ -46,7 +46,7 @@ export type EventMap = {
     numberOfValuesFiles: number;
     executionTime: number;
   };
-  'app_start/create_project': {from: 'scratch' | 'git' | 'template' | 'folder'; templateID?: string};
+  'app_start/create_project': {from: 'sample' | 'scratch' | 'git' | 'template' | 'folder'; templateID?: string};
```
olensmar, line 49: add `helm` option?

olensmar (approved):
thanks!

devcatalin:
Added the events for #4097 and #4093

devcatalin:
Also added telemetry for #4098

src/editor/editor.instance.ts
```diff
@@ -168,6 +172,7 @@ export const clearEditorDecorations = () => {
 
 monaco.languages.registerHoverProvider('yaml', {
   provideHover: (model, position) => {
+    trackEvent('editor/hover');
```
olensmar, line 175: perhaps we should track this after checking if we actually have something to show?

src/shared/models/telemetry.ts
```diff
@@ -117,6 +123,7 @@ export type EventMap = {
   'cluster/actions/scale': {replicasNumber: number};
   'cluster/actions/restart': undefined;
   'cluster/actions/delete': {kind: string};
+  'cluster/metrics': {providers: string[]};
```
olensmar, line 126: can we add the k8s version and runtime? (which we show in the dashboard)

src/editor/editor.instance.ts
devcatalin: Since Monaco could provide it's own hovers, for YAML / Kubernetes schema errors, which are not created by us, I it makes more sense to send that event for any hover.
I added a comment about this to the code

olensmar: agree - I thought this method was called by Monaco for any hover opportunity - it's first after the editorHovers.filter that we know if there is actually something to show.. !?

devcatalin: I added a "type" property for our own hovers and now check if I can send that or not

olensmar (dismissed):
awesome - thank you!

olensmar (approved):

PR #4216, opened 2024-02-26
chore(release): 2.4.5

monojack:
This PR...

## Changes

-

## Fixes

-

## How to test it

-

## Screenshots

-

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test


olensmar:
@monojack could we check if there are any major electron/electron-builder upgrades available?

monojack:
> @monojack could we check if there are any major electron/electron-builder upgrades available?

Electron has 2 major increments. I took a quick look at the documented breaking changes and tried a search through the source code - looks like we're not using any of the changed/deprecated/removed symbols. 

image

olensmar:
thanks @monojack - perhaps there is a 27.X upgrade that we should use for security fixes that won't break our build?

monojack:
Was just checking 27.x and 28.x, they both work. Which one would you like??

olensmar:
then I'd go for 28 - thanks!

olensmar (approved):

